Issue #16, opened 2020-06-30
Add Stan Segmentation Example


blaiszik:
Dataset:  _test_foundry_stan_dendrite_segmentation_v1.1

@ethantruelove

PR #17, opened 2020-06-30
Additional example cleaning

ethantruelove:
Removed miscellaneous data files, images/ examples. Added in dendrite-segmentation example. Renamed bg_data to bandcaps for consistency.


blaiszik:
Closes #16

PR #60, opened 2021-03-01
Fixing XTract crawling

blaiszik:
This commit addresses 
* Usage of return values from the new XTract API, 
* Fixes an authentication issue where incorrect tokens were being passed to funcX

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -53,6 +53,7 @@ class Foundry(FoundryMetadata):
                 "petrel",
                 "transfer",
                 "dlhub",
+                "openid",
                 "https://auth.globus.org/scopes/facd7ccc-c5f4-42aa-916b-a0e270e2c2a9/all",
             ],
             app_name="Foundry",
@@ -82,7 +83,7 @@ class Foundry(FoundryMetadata):
         self.xtract_tokens = {
             "auth_token": auths["petrel"].access_token,
             "transfer_token": auths["transfer"].authorizer.access_token,
-            "funx_token": auths[
+            "funcx_token": auths[
                 "https://auth.globus.org/scopes/facd7ccc-c5f4-42aa-916b-a0e270e2c2a9/all"
             ].access_token,
         }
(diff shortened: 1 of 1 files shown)


foundry/foundry.py
```diff
@@ -380,18 +381,19 @@ def download(self, globus=True, verbose=False, **kwargs):
             crawl_url = f"{xtract_base_url}/crawl"
             if verbose:
                 print(f"Crawl URL is : {crawl_url}")
+
+            first_ep_dict = {
```
ascourtas, line 385: what does "first ep" refer to? I see in a comment above it's some kind of crawling service? oh so is this the first set of search criteria for the crawler (with the idea that there will be other queries in the future)?

foundry/foundry.py
```diff
@@ -380,18 +381,19 @@ def download(self, globus=True, verbose=False, **kwargs):
...
-                    "repo_type": "GLOBUS",
-                    "eid": source_ep_id,
-                    "dir_path": folder_to_crawl,
-                    "Transfer": transfer_token,
-                    "Authorization": funcx_token,
-                    "grouper": grouper,
-                    "https_info": {"base_url": base_url},
-                },
+                f"{xtract_base_url}/crawl",
+                json={"endpoints": [first_ep_dict], "tokens": tokens},
```
ascourtas, line 394: I'm surprised the EP query is described as an "endpoint", could you clarify this for my understanding?

ascourtas (commented):
basically just have a couple of clarifying questions for my own understanding!

foundry/foundry.py
blaiszik: This is terminology from Tyler, I am not sure why "first" is descriptive here as there is no "second" anywhere. I'll change this to `ep_dict`

foundry/foundry.py
blaiszik: Maybe we are overloading the term endpoint. Endpoint can be used to describe where a query is being sent to an API, but also can refer to a Globus endpoint. Here we are sending the JSON payload in the variable `json` to the `crawl` _API endpoint_ which requires knowing which _Globus endpoint_ to crawl. Does that help clarify? Pretty confusing!

foundry/foundry.py
ascourtas: thanks Ben!

foundry/foundry.py
ascourtas: yes that makes more sense! do you think this would warrant an explanatory comment? for future devs

PR #61, opened 2021-03-02
documenting how to build with Sphinx

ascourtas:
basic instructions for developers on how to build documentation locally. also brief pointer to WIP docs on Read the Docs


ascourtas:
would love to know if ya'll think the README is the right place for developer instructions!

kjschmidt913:
> would love to know if ya'll think the README is the right place for developer instructions!


Do you think it would make sense to make a readme inside the docs folder? That way the readme on the main page isn't quite as long. Or do you think that makes it harder to find?

ascourtas:
> > would love to know if ya'll think the README is the right place for developer instructions!
> 
> Do you think it would make sense to make a readme inside the docs folder? That way the readme on the main page isn't quite as long. Or do you think that makes it harder to find?

@kjschmidt913  hmm yeah my only concern with that is it’d be hard to find. If I’m just looking at the repo instructions I wouldn’t initially think to look in the /docs folder. But maybe that’s ok?

blaiszik:
As for where to put content, maybe its worth looking at a few major repos and seeing how they handle it. Keras, Scikit-Learn, etc.

kjschmidt913:
Okay! I was thinking if someone was looking for how to contribute to/update the docs that they may look in the docs folder since it's visible on the repo page. I've seen people use readme's in different folders throughout repos and have found it helpful. But maybe this makes more sense on the main readme, especially because it isn't that long. I was concerned that the main readme would get long/overwhelming but it's not bad at all!

kjschmidt913:
Oh sorry, Ben, I think I was typing that as you posted. Good idea to look around and see what other repos looks like. That will be very helpful

PR #67, opened 2021-03-15
Data package spec

blaiszik:


ascourtas:
@blaiszik is this ready for review? If so I think the import and package spec updates look good! can't tell what changed in the notebook though, maybe @kjschmidt913 could speak more to that? just before I approve it

ascourtas (approved):
looks good!

PR #68, opened 2021-03-17
Aristanascourtas/ch266/wrap mdf connect client method for publishing

ascourtas:
- amend and expand `publish()` for Foundry, such that MDF methods are abstracted away
- add test for `publish()`
- still needs some additional test handling, notably dealing with curation and cleanup after test publications


foundry/foundry.py
```diff
@@ -252,36 +262,68 @@ def describe(self):
         print("DC:{}".format(self.dc))
         print("Dataset:{}".format(self.dataset.json(exclude={"dataframe"})))
 
-    def publish(self, foundry_metadata, update=False, **kwargs):
+    def publish(self, foundry_metadata, data_source, title, authors, update=False, publication_year=None, **kwargs):
```
ascourtas, line 265: I made it so the metadata, data_source, title, and authors were all required arguments. Does that sound right to ya'll?

requirements.txt
```diff
@@ -6,6 +6,8 @@ dlhub_sdk>=0.0.1
 h5py>=2.10.0
 pandas==1.2.2
 pydantic>=1.6.1
-sphinx-materialdesign-theme==0.1.11
+Sphinx>=3.5.1
```
ascourtas, line 9: fixed an omission I had made in an earlier PR -- forgot to update requirements.txt!

PR #72, opened 2021-03-17
add new publication example notebook

ascourtas:
- notebook covers how to publish using new `f.publish()` method
- MDF is abstracted away
- requires newest version of `foundry_ml` from PyPi


blaiszik:
1. `title` is generally more of a free form string, perhaps change the example to something like "Scourtas Example Iris Dataset"
2. Do we need to import `import mdf_connect_client`?
3. I would clear the output from the cell with the installs. It is just a bit tedious to scroll past.
4. In order to publish, the user has to first join a group, add a link to this group: https://app.globus.org/groups/cc192dca-3751-11e8-90c1-0a7c735d220a/about

ascourtas:
> 1. `title` is generally more of a free form string, perhaps change the example to something like "Scourtas Example Iris Dataset"
> 2. Do we need to import `import mdf_connect_client`?
> 3. I would clear the output from the cell with the installs. It is just a bit tedious to scroll past.
> 4. In order to publish, the user has to first join a group, add a link to this group: https://app.globus.org/groups/cc192dca-3751-11e8-90c1-0a7c735d220a/about

thanks @blaiszik ! I'll make the edits

PR #74, opened 2021-03-18
Add requirements.txt file for examples

blaiszik:


ascourtas (approved):
looks good to me! we had slack convos about data schema edits already. I like the comments added to the notebook example

PR #75, opened 2021-03-18
Data package build

blaiszik:


foundry/foundry.py
```diff
@@ -498,3 +512,41 @@ def download_file(file):
...
+
+        if file:
+            with open(file, "r") as fp:
+                fs = FoundrySpecification(**json.load(fp))
+        else:
+            fs = FoundrySpecification(**spec)
+
+        fs.remove_duplicate_dependencies()
+
+        results = Parallel(n_jobs=num_cores)(
```
ascourtas, line 546: what does this do? can't tell from context, could you please add an inline comment?

foundry/foundry.py
```diff
@@ -498,3 +512,41 @@ def download_file(file):
...
+        Returns
+        -------
+        (Foundry): self: for chaining
+        """
+
+        print("Building Data Package")
+        num_cores = multiprocessing.cpu_count()
+
+        def start_download(ds, interval=interval, globus=False):
+            print("=== Fetching Data Package {} ===".format(ds.name))
```
ascourtas, line 533: I like the progress print statements, v helpful!

foundry/foundry.py
```diff
@@ -498,3 +512,41 @@ def download_file(file):
             print(results)
 
         return self
+
+    def build(self, spec, globus=False, interval=3, file=False):
+        """Build a Foundry Data Package
```
ascourtas, line 517: Could you add a little more clarification in the docstring on how this is used? Looks like it instantiates the foundry data package specification but I'm not entirely sure how that would be used.

ascourtas: after spending some more time on this function, it looks like it actually downloads the data package per the FoundrySpec? is that right?

ascourtas (changes requested):
just a few clarification requests on `build()`, for the most part. weird that some of the edits are exactly the same as in the previous PR we just merged, should be ok though

foundry/foundry.py
```diff
@@ -293,7 +307,7 @@ def publish(self, foundry_metadata, data_source, title, authors, update=False, p
             affiliations=kwargs.get("affiliations", []),
             subjects=kwargs.get("tags", ["machine learning", "foundry"]),
             publisher=kwargs.get("publisher", ""),
-            publication_year=publication_year
+            publication_year=publication_year,
```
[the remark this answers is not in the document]
blaiszik: This is an auto formatting thing from Black formatter. I'm not sure exactly why it does this, but it does it automatically...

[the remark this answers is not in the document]
ascourtas: I googled it, somehow it's supposed to be useful when diffing files in version control? https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas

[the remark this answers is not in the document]
ascourtas: I'm fine with leaving it!

PR #76, opened 2021-03-18
Doi spec support

blaiszik:


examples/bandgap/bandgap_demo.ipynb
```diff
@@ -25,228 +25,55 @@
...
-      "Done curling.\n",
-      "[{'/foundry/_test_foundry_assorted_computational_bandgap_v1.1/foundry_dataframe.json status': True}]\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
-    "for p in requirements['dependencies']:\n",
-    "    f = f.load(p, globus=False, verbose=True)"
+    "from foundry import Foundry\n",
```
ascourtas, line 74: this looks potentially different from your other two PRs? maybe just double check before we merge this that this notebook is gonna look the way you want it to?

foundry/foundry.py
```diff
@@ -87,7 +92,16 @@ def __init__(
             ].access_token,
         }
 
-    def load(self, name, download=True, globus=True, verbose=False, **kwargs):
+    def load(
```
ascourtas, line 95: I believe this is not PEP 8 compliant https://www.python.org/dev/peps/pep-0008/#indentation

foundry/foundry.py
```diff
@@ -100,19 +114,35 @@ def load(self, name, download=True, globus=True, verbose=False, **kwargs):
...
             self
         """
+        res = []
         # MDF specific logic
-        res = self.forge_client.match_field(
-            "mdf.organizations", "foundry"
-        ).match_resource_types("dataset")
-        res = res.match_field("mdf.source_id", name).search()
+        # Handle DOI inputs
+        if name.startswith("10.") and provider == "MDF":
```
ascourtas, line 120: will DOIs always start with "10"?

foundry/foundry.py
```diff
@@ -100,19 +114,35 @@ def load(self, name, download=True, globus=True, verbose=False, **kwargs):
...
+            )
 
-        # TODO: if object empty, handle
-        res = res[0]
-        res["dataset"] = res["projects"]["foundry"]
-        res["dataset"]["type"] = res["dataset"]["package_type"]
-        del res["projects"]["foundry"]
+        # Handle MDF source_ids
+        else:
+            print("source_id")
```
ascourtas, line 130: did you want to also print the actual source_id here?

foundry/foundry.py
```diff
@@ -100,19 +114,35 @@ def load(self, name, download=True, globus=True, verbose=False, **kwargs):
...
+                .search()
+            )
 
-        self = Foundry(**res)
+        if not res:
+            return self
+        else:
+            res = res[0]
+            res["dataset"] = res["projects"]["foundry"]
+            res["dataset"]["type"] = res["dataset"]["package_type"]
```
ascourtas, line 143: for my own learning, could you explain what's going on here? why would we need to restructure the result like this?

foundry/foundry.py
```diff
@@ -100,19 +114,35 @@ def load(self, name, download=True, globus=True, verbose=False, **kwargs):
...
+            del res["projects"]["foundry"]
+            self = Foundry(**res)
```
ascourtas, line 145: also unclear to me what's going on here, we replace the existing instance??

ascourtas (changes requested):
same thing as before where we have repeated commits in this PR. my guess is when you're making local branches, they're branching off each other? I'm just ignoring the edits that belong to other PR's and that I've already commented on.

mostly just asking for clarification on things.

foundry/foundry.py
blaiszik: Yes, DOIs always start with the string `10. I have no idea why they do that though...

foundry/foundry.py
blaiszik: This is the Python Black autoformatting. It's ok to accept this for now, but I will look for a friendlier code formatter

foundry/foundry.py
blaiszik: I had to think on this a bit too...which means I should be adding more comments :) 

Lines 141 onward map the Forge return value into the FoundryMetadata Pydantic dictionary. The def for the Pydantic is in models.py

foundry/foundry.py
blaiszik: Yes, I think this was because when I was trying to re-set values within the self object they weren't sticking. This is probably not the best way to do it...

ascourtas: gotcha, my concern here is that this may result in some unexpected behavior and tricky bugs down the road. I might just wanna flag it for potential re-working in the future.

foundry/foundry.py
ascourtas: adding clarifying print statements

foundry/foundry.py
ascourtas: adding comment

foundry/foundry.py
ascourtas: I'll add comments/TODOs

PR #87, opened 2021-04-30
Nb exmaples

kjschmidt913:
- I've updated all of the notebooks with tutorial-like language (moreso on the beginning steps of the notebook since those are the biggest hurdles for people getting started with foundry!)
- Each notebook has a block of code that detects if the user is running in colab or locally
- There is a requirements.txt file for each notebook in its folder for users running the notebook locally
- I'm still having an issue with the matminer install. I know Ben saw that they fixed this bug but just hadn't released the fix yet, so I'm just hoping they do that soon. If you have a suggestion for a fix before then, please let me know or just add it yourself.
- Besides the matminer issue in the bandgap notebook, all other notebooks are able to run smoothly


examples/bandgap/bandgap_demo.ipynb
```diff
@@ -1,1001 +1,637 @@
...
-     ]
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "kernelspec": {
+      "display_name": "Python 3",
+      "language": "python",
+      "name": "python3"
+    },
+    "language_info": {
```
ascourtas, line 10: Comment for `dendrite_segmentation` notebook because Github won't let me comment on it directly (diff too big)
---------------------------------------------------------------

I'd add one sentence to the *Create Metadata* section, maybe just:

Create the metadata for this dataset so you can push it to Foundry if you like (see other notebooks for dataset publication).

examples/iris/iris.ipynb
```diff
@@ -1,323 +1,503 @@
...
-   "cell_type": "code",
-   "execution_count": 2,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "#need Globus to pull data\n",
-    "f = Foundry().load(\"_test_blaiszik_foundry_iris_v2.1\")\n",
-    "X,y = f.load_data()"
-   ]
+  "nbformat": 4,
```
ascourtas, line 2: At the end, I think the title *Only Run to Create Local Metadata and Dataframe* should be changed to *Create Local Metadata and Dataframe* and then a sentence should be added below that says something like: 
Shape your data (in this case, the Iris dataset from Scikit-Learn) into a dataframe and save it as JSON if you want to publish it to Foundry later (see other notebooks on how to publish datasets). Save the file to 'foundry_dataframe.json'

Similarly, you need to create a metadata file for your dataset to tell Foundry what your data is all about! Format your metadata like so, and then save it to 'foundry_metadata.json'

examples/oqmd/oqmd.ipynb
```diff
@@ -1,670 +1,777 @@
...
-   "cell_type": "code",
-   "execution_count": 3,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "#need Globus to pull data\n",
-    "f = f.load('_test_foundry_oqmd_v1.1')\n",
-    "X,y = f.load_data()"
-   ]
+  "nbformat": 4,
```
ascourtas, line 2: Instead of *Run to Create Dataframe and Metadata Locally without Fetch* I think it should be something like *Create Dataframe and Metadata Locally*, with a sentence or two like what I suggested for iris.ipynb

examples/oqmd/oqmd.ipynb
```diff
@@ -1,670 +1,777 @@
...
+  "nbformat_minor": 0,
+  "metadata": {
```
ascourtas, line 4: Does this notebook run properly, it looks like there's a reference to a function called `oqmd_to_df()` that's never defined, nor is it imported?? I have no idea what's going on here

ascourtas (changes requested):
some comments to add content to the ML sections, some comments on having an easy set of foundry setup commands for the user, also concern that maybe the oqmd notebook doesn't work?

examples/oqmd/oqmd.ipynb
kjschmidt913: found the missing cell! Great catch

kjschmidt913:
I've made all the changes. Thanks for the comments for the ML sections and the OQMD notebook catch! I didn't notice I had accidentally deleted that cell. V important. They should be good now. I've committed all teh changes to this branch so should be good to merge

examples/oqmd/oqmd.ipynb
```diff
@@ -368,7 +368,8 @@
         "id": "wTKDm0BXgdfg"
       },
       "source": [
-        "# Run to Create Dataframe and Metadata Locally without Fetch"
+        "# Create Local Metadata and Dataframe\n",
+        "To publish your dataset to Foundry, you'll need to shape the data into a dataframe and save it as JSON, which you can do by running the code below. The file needs to be saved as `foundry_dataframe.json`. You'll also need to create a metadata file for your dataset so Foundry knows what your data is all about! Format your metadata using the code below, and then save it to `foundry_metadata.json`."
```
ascourtas, line 372: love this

ascourtas (approved):
this looks great <3

PR #103, opened 2021-06-22
work on load function

zk794:
added error handling, including empty dataset name

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -133,22 +133,16 @@ class Foundry(FoundryMetadata):
             res = res[0] # if search returns multiple results, this automatically uses first result
         except IndexError as e:
             raise Exception("load: No metadata found for given dataset") from e
-        except BaseException as e:
-            raise Exception("load: An unknown error occurred") from e
 
         try:
             res["dataset"] = res["projects"][self.config.metadata_key]
         except KeyError as e:
-            raise Exception("load: Problem with projects field") from e
-        except BaseException as e:
-            raise Exception("load: An unknown error occurred") from e
+            raise Exception("load: not able to index with metadata key {}".format(self.config.metadata_key)) from e
 
         try:
             res["dataset"]["type"] = res["dataset"]["data_type"]
         except KeyError as e:
             raise Exception("load: dataset field does not have data_type") from e
-        except BaseException as e:
-            raise Exception("load: An unknown error occurred") from e
 
         del res["projects"][self.config.metadata_key]
 
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -122,10 +122,36 @@ class Foundry(FoundryMetadata):
         else:
             res = metadata
 
-        # TODO: if object empty, handle
-        res = res[0]
-        res["dataset"] = res["projects"][self.config.metadata_key]
-        res["dataset"]["type"] = res["dataset"]["data_type"]
+        # we figured from our tests that at each of these three lines there could be an error
+        # if res is an empty list. (we also see that res will always be an empty list,
+        # never None or not a list)
+        try:
+            res = res[0]
+        except IndexError:
+            print("No metadata found for given dataset")
+            exit()
+        except:
+            print("An unknown error occurred")
+            exit()
+
+        try:
+            res["dataset"] = res["projects"][self.config.metadata_key]
+        except KeyError:
+            print("Problem with projects field")
+            exit()
+        except:
+            print("An unknown error occurred")
+            exit()
+
+        try:
+            res["dataset"]["type"] = res["dataset"]["data_type"]
+        except KeyError:
+            print("dataset field does not have data_type")
+            exit()
+        except:
+            print("An unknown error occurred")
+            exit()
+
         del res["projects"][self.config.metadata_key]
 
         self = Foundry(**res)
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -139,11 +139,6 @@ class Foundry(FoundryMetadata):
         except KeyError as e:
             raise Exception("load: not able to index with metadata key {}".format(self.config.metadata_key)) from e
 
-        try:
-            res["dataset"]["type"] = res["dataset"]["data_type"]
-        except KeyError as e:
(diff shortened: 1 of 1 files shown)


foundry/foundry.py
```diff
@@ -113,6 +113,11 @@ def load(self, name, download=True, globus=True, verbose=False, metadata=None, *
         -------
             self
         """
+        # handle empty dataset name (was returning all the datasets)
+        if not name:
+            print("load: No dataset name is given")
+            exit()
```
ascourtas, line 119: instead of `exit()`, we should `raise` an Exception. See https://docs.python.org/3/tutorial/errors.html#raising-exceptions for how to properly raise an Exception. Here is the list of Python Exceptions (you want to pick the one that is the most specific, and makes sense for the issue): https://docs.python.org/3/library/exceptions.html

ascourtas (changes requested):

foundry/foundry.py
zk794: I made a change!

foundry/foundry.py
```diff
@@ -122,10 +126,30 @@ def load(self, name, download=True, globus=True, verbose=False, metadata=None, *
...
-        res = res[0]
-        res["dataset"] = res["projects"][self.config.metadata_key]
-        res["dataset"]["type"] = res["dataset"]["data_type"]
+        # we figured from our tests that at each of these three lines there could be an error
+        # if res is an empty list. (we also see that res will always be an empty list,
+        # never None or not a list)
+        try:
+            res = res[0] # if search returns multiple results, this automatically uses first result
+        except IndexError as e:
+            raise Exception("load: No metadata found for given dataset") from e
```
ascourtas, line 135: this is great, nice and informative!

foundry/foundry.py
```diff
@@ -122,10 +126,30 @@ def load(self, name, download=True, globus=True, verbose=False, metadata=None, *
...
+        except BaseException as e:
+            raise Exception("load: An unknown error occurred") from e
```
ascourtas, line 137: you don't need to except the `BaseException` each time, since that'll automatically be raised and won't be caught by the `IndexError`. And we don't have any more information to give the user for this unknown exception (they will already know it happened in the `load()` function because it'll be part of the stack trace)

foundry/foundry.py
```diff
@@ -122,10 +126,30 @@ def load(self, name, download=True, globus=True, verbose=False, metadata=None, *
...
+
+        try:
+            res["dataset"] = res["projects"][self.config.metadata_key]
```
ascourtas, line 140: this could be an issue with "projects", or it could be an issue with the metadata_key. I would suggest changing the message here to something like `"not able to index with metadata key {}".format(self.config.metadata_key)`

ascourtas (changes requested):

ascourtas:
@blaiszik I believe the assignment operator was already how the code was written (without any try/except clauses). 

now that I think about it, what exactly is the purpose of the line `res["dataset"]["type"] = res["dataset"]["data_type"]`? what is the purpose of overwriting "type" with "data_type"? I ask so we can figure out what's the most straightforward

blaiszik:
> @blaiszik I believe the assignment operator was already how the code was written (without any try/except clauses).
> 
> now that I think about it, what exactly is the purpose of the line `res["dataset"]["type"] = res["dataset"]["data_type"]`? what is the purpose of overwriting "type" with "data_type"? I ask so we can figure out what's the most straightforward

I suspect this is an attempt at compatibility between our two metadata schemas. If `type` has been replaced by `data_type` throughout, we should be compatible with all the new schemas (which I think is all we care about now)

ascourtas:
> > @blaiszik I believe the assignment operator was already how the code was written (without any try/except clauses).
> > now that I think about it, what exactly is the purpose of the line `res["dataset"]["type"] = res["dataset"]["data_type"]`? what is the purpose of overwriting "type" with "data_type"? I ask so we can figure out what's the most straightforward
> 
> I suspect this is an attempt at compatibility between our two metadata schemas. If `type` has been replaced by `data_type` throughout, we should be compatible with all the new schemas (which I think is all we care about now)

awesome, @zk794 @ribhavb can you test what happens when you don't have this line and see if `res` still looks the way it's supposed to? if so, we can delete it altogether and have fewer lines of code :)

ribhavb:
@ascourtas When running the code and looking at the res object without the line, it looks like it contains everything and the code runs normal

ribhavb:
@ascourtas Actually, on running some more tests, it seems like the `res["dataset"]["type"] = res["dataset"]["data_type"]` is needed because in the _load_data() function, for the line: 
`if self.dataset.type.value == "tabular":`, we always get an attribute error because there is no type

We get error: 
`FAILED test_foundry.py::test_dataframe_load - AttributeError: 'NoneType' object has no attribute 'value'`

ascourtas:
> @ascourtas Actually, on running some more tests, it seems like the `res["dataset"]["type"] = res["dataset"]["data_type"]` is needed because in the _load_data() function, for the line:
> `if self.dataset.type.value == "tabular":`, we always get an attribute error because there is no type
> 
> We get error:
> `FAILED test_foundry.py::test_dataframe_load - AttributeError: 'NoneType' object has no attribute 'value'`

thanks for investigating @ribhavb ! part of the issue there is that test needs to be updated to check for `self.dataset.data_type.value` instead of `self.dataset.data.value` -- this was part of the recent metadata schema changes I mentioned. I just created a new Clubhouse story to update the tests. So truly we shouldn't (in theory) need to assign anything to `res["dataset"]["type"]`, assuming the datasets are all updated to the most recent metadata format (which they need to be anyways). I think it's ok to axe this line altogether, but maybe merge this into `dev` instead of into `main`. What do you think @blaiszik ?

ascourtas:
FYI you can see what the most recent metadata JSON schema is supposed to be here: https://github.com/materials-data-facility/data-schemas/blob/master/schemas/projects.json#L111

ascourtas:
checked in with Ben re: that one line, looks good!

ascourtas (approved):

PR #105, opened 2021-06-24
add read_csv logic

zk794:


zk794:
worked on [can't read dataframe file story](https://app.clubhouse.io/ai-for-materials-and-chemistry/story/493/more-descriptive-error-for-no-file) here as well as the csv support one because they both require work in the same area and therefore seem related

foundry/foundry.py
```diff
@@ -725,10 +725,19 @@ def _load_data(self, file=None, source_id=None, globus=True):
                 )
             except:
                 # Try to read individual lines instead
-                self.dataset.dataframe = pd.read_json(
-                    os.path.join(path, file), lines=True
-                )
-
+                try:
```
ascourtas, line 728: hm I thought I already commented on this but idk what happened to it -- did ya'll see my comment about un-nesting this so it's a chain of try/except? as shown in https://docs.python.org/3/tutorial/errors.html#handling-exceptions

ascourtas (changes requested):

foundry/foundry.py
zk794: We didn't see any comment. We were wondering if there is another way to do this though. Looking at that page, it is still unclear to me how else to tell it to try three things and only if they all fail then raise the error

ascourtas: ah I see, to do that you'd put them all in the same `try` block. like 

```
try: 
    thing 1
    thing 2 
    thing 3
except IOError as e:
    do stuff if IO error is thrown by 1 of those 3 lines
except ValueError as e:
    do stuff if value error is thrown by 1 of those 3 lines,  etc
```

ascourtas: If that doesn't make sense lmk and we can hop on a call!

PR #109, opened 2021-07-07
Fixes to test_foundry.py

ribhavb:
Updated metadata for publishing test; script to delete downloaded test data in relevant tests; checked that tests work on local machine with success


ribhavb:
Also for test_data_source variable - changed the destination to the foundry_test folder that I found with the globus file manager -- not 100% sure if this was the correct thing to do, but the publish tests were passing

PR #112, opened 2021-07-08
Csv_read functionality

ribhavb:
added ability to try read_csv after trying to read file as json.

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -36,7 +36,6 @@ class Foundry(FoundryMetadata):
     TODO:
     -------
     Add Docstring
-
     """
 
     # transfer_client: Any
@@ -106,10 +105,8 @@ class Foundry(FoundryMetadata):
             globus (bool): If True, download using Globus, otherwise https
             verbose (bool): If True print additional debug information
             metadata (dict): **For debug purposes.** A search result analog to prepopulate metadata.
-
         Keyword Args:
             interval (int): How often to poll Globus to check if transfers are complete
-
         Returns
         -------
             self
@@ -153,7 +150,6 @@ class Foundry(FoundryMetadata):
 
     def list(self):
         """List available Foundry data packages
-
         Returns
         -------
             (pandas.DataFrame): DataFrame with summary list of Foundry data packages including name, title, and publication year
@@ -178,10 +174,8 @@ class Foundry(FoundryMetadata):
 
     def get_packages(self, paths=False):
         """Get available local data packages
-
         Args:
            paths (bool): If True return paths in addition to package, if False return package name only
-
         Returns
         -------
             (list): List describing local Foundry packages
@@ -198,7 +192,6 @@ class Foundry(FoundryMetadata):
         """Collect dataframes of local data packages
         Args:
            packages (list): List of packages to collect, defaults to all
-
         Returns
         -------
             (tuple): Tuple of X(pandas.DataFrame), y(pandas.DataFrame)
@@ -222,15 +215,12 @@ class Foundry(FoundryMetadata):
 
     def run(self, name, inputs, **kwargs):
         """Run a model on data
-
         Args:
            name (str): DLHub model name
            inputs: Data to send to DLHub as inputs (should be JSON serializable)
-
         Returns
         -------
              Returns results after invocation via the DLHub service
-
         TODO:
         -------
         - Pass **kwargs through to DLHub client and document kwargs
@@ -239,19 +229,14 @@ class Foundry(FoundryMetadata):
 
     def load_data(self, source_id=None, globus=True):
         """Load in the data associated with the prescribed dataset
-
         Tabular Data Type: Data are arranged in a standard data frame
         stored in self.dataframe_file. The contents are read, and
-
         File Data Type: <<Add desc>>
-
         For more complicated data structures, users should
         subclass Foundry and override the load_data function
-
         Args:
            inputs (list): List of strings for input columns
            targets (list): List of strings for output columns
-
         Returns
         -------s
              (tuple): Tuple of X, y values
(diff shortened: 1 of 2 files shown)


tests/test_foundry.py
```diff
@@ -10,9 +10,12 @@
 from dlhub_sdk import DLHubClient
 from mdf_connect_client import MDFConnectClient
 
-test_dataset = "_test_blaiszik_foundry_iris_v2.1"
-expected_title = "Foundry - Iris Dataset"
-test_metadata = {
+#updated test dataset
+test_dataset = "_test_foundry_iris_dev_v2.1"
```
ascourtas, line 14: good update

tests/test_foundry.py
```diff
@@ -24,8 +27,60 @@
...
+    "data_type": "tabular",
+    'task_type': ['unsupervised', 'generative'],
+    'domain': ['materials science', 'chemistry'],
+    'n_items': 1000
+}
+# Globus endpoint for '_iris_dev'
+test_data_source = "https://app.globus.org/file-manager?origin_id=e38ee745-6d04-11e5-ba46-22000b92c6ec&origin_path=%2Ffoundry-test%2F"
+
+#Quick function to delete any downloaded test data
+def delete_test_data(f):
```
ascourtas, line 79: great, I would just rename this to `_delete_test_data()` so the user knows it's meant to be a private function.

also, could you make `f` a little more informative? unclear if it takes a folder path or a filename

tests/test_foundry.py
```diff
@@ -61,39 +116,56 @@ def test_list():
...
+
     f = Foundry(no_browser=True, no_local_server=True)
-    f = f.load(test_dataset, download=True)
-    X, y = f.load_data()
+
+    delete_test_data(f)
+
+    f = f.load(test_dataset, download=True, globus=False)
+    res = f.load_data()
+    X, y = res['train']
```
ascourtas, line 160: good fixes

foundry/foundry.py
```diff
@@ -618,12 +618,25 @@ def _load_data(self, file=None, source_id=None, globus=True):
...
                     os.path.join(path, file)
                 )
-            except:
-                # Try to read individual lines instead
-                self.dataset.dataframe = pd.read_json(
-                    os.path.join(path, file), lines=True
-                )
-
+            except Exception:
+                pass
```
ascourtas, line 622: you shouldn't need to pass on any of these exceptions, if they are raised you'll want them to bubble up. instead do the try/except chaining outlined in https://docs.python.org/3/tutorial/errors.html#handling-exceptions

tests/test_foundry.py
```diff
@@ -24,8 +27,60 @@
     "short_name": "iris_example",
     "package_type": "tabular"
 }
-# Globus endpoint for '_test_blaiszik_foundry_iris_v2.1'
-test_data_source = "https://app.globus.org/file-manager?origin_id=e38ee745-6d04-11e5-ba46-22000b92c6ec&origin_path=%2Ffoundry%2F"
+
+
+test_metadata = {
```
ascourtas, line 32: metadata looks good! 👍

ascourtas (changes requested):

ascourtas:
@ribhavb @ZKatok @zk794 could you compare this to the other CSV read PR and see whether that other one can be closed?

also, there are changes in this PR for general test-infrastructure (such as deleting old data and what not). is that in any other PR? if so, please clean up so it's only in one PR :)

ribhavb:
Makes sense! I think the other CSV read PR had already been closed from last week . 
My bad with the test changes -- I closed the other PR I made for it.

ribhavb:
@ascourtas  Updated the code-- printed the errors in csv and made the change to the data deletion function

foundry/foundry.py
```diff
@@ -618,12 +618,26 @@ def _load_data(self, file=None, source_id=None, globus=True):
...
-            except:
-                # Try to read individual lines instead
-                self.dataset.dataframe = pd.read_json(
-                    os.path.join(path, file), lines=True
-                )
-
+            except Exception as e:
+                print("Reading {} as JSON failed: {} \n".format(file, e), "Now attempting to read as JSONL")
+
+            if (self.dataset.dataframe is None):
```
ascourtas, line 624: currently, this code will try to read the file each of these ways every time, which we dont necessarily want -- if it works one way, we want to stop trying

ascourtas: so instead of having separate blocks, I think this can all be condensed into slightly cleaner code by doing some sort of exception chaining? I'm trying to think of the best way to do it, but I think it'd require a little playing around here. something like 
```
try: 
    one way
    except:
         print that this way failed, what you're about to try
         try:
               second way
         except:
               print that second way failed
               try:
                    etc
```
I believe you or Zoa originally had this in the other PR, and I suggested one `try` with multiple `excepts`, but after further reading I realize that doesn't actually work as I expected (my bad!). 

also, why are you checking `if (self.dataset.dataframe is None)`, could you clarify the logic here?

ascourtas: something like this: https://stackoverflow.com/questions/17322208/multiple-try-codes-in-one-block

ribhavb: @ascourtas, the `if (self.dataset.dataframe is None)` is for that, if one of the readings works, so if it is read as a JSON successfully, then the statement evaluates to false and doesn't attempt any of the other readings -- I was able to check this in testing and it was working as intended, so when a certain type of reading fails the dataframe is still none, but once it works it is no longer none and doesnt attempt another read-- but we can write the nesting way as earlier if that would be better

ascourtas: @ribhavb ahhh I see! that makes sense. I feel like there might be a way to make it more concise, but I cant think of it off the top of my head. I say this is good for now, and we can always refactor in the future if we wanna

ascourtas:
looks good!

PR #113, opened 2021-07-08
Add capability to accept external authorizers

blaiszik:
Add Forge init argument that allows for accepting external authorizers. Users can then authenticate with confidential clients and pass these authorizers to Forge rather than relying on the built-in authentication. This should enable full CI/CD testing for Foundry.


ascourtas (approved):
looks good to me!

PR #114, opened 2021-07-12
first improper download commit

ribhavb:
For clubhouse task: https://app.clubhouse.io/ai-for-materials-and-chemistry/story/490/handle-dataframe-not-downloading

Code at beginning of method is to make sure that in the case where the directory exists, but doesn't contain all the files, it will attempt to redownload the dataset (last week when trying to test stuff ran into this issue where this issue would cause errors further down the road) 

Code at the end makes sure that directory exists and all files are present.

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -27,7 +27,7 @@ import json
 import glob
 import h5py
 import time
-import os
+import os, shutil
 from foundry.xtract_method import *
 
 
@@ -504,8 +504,17 @@ class Foundry(FoundryMetadata):
         """
         # Check if the dir already exists
         path = os.path.join(self.config.local_cache_dir, self.mdf["source_id"])
+
         if os.path.isdir(path):
-            return self
+            dir_length = len(os.listdir(path))
+            if self.dataset.splits:
+                # if metadata indicates splits, check that the directory has as many files as there are splits
+                if(dir_length == len(self.dataset.splits)):
+                    return self
+            else:
+                # in the case of no splits, ensure the directory contains the data file
+                if(dir_length == 1):
+                    return self
 
         res = self.forge_client.search(
             "mdf.source_id:{name}".format(name=self.mdf["source_id"]), advanced=True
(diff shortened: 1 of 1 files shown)


foundry/foundry.py
```diff
@@ -504,8 +504,17 @@ def download(self, globus=True, verbose=False, **kwargs):
         """
         # Check if the dir already exists
         path = os.path.join(self.config.local_cache_dir, self.mdf["source_id"])
+
         if os.path.isdir(path):
```
ascourtas, line 508: I would make it clear (either with a comment or with an else statement, if there are no unintended consequences) that if the dir is present but the incorrect number of files are in it, then the download would be re-attempted

foundry/foundry.py
```diff
@@ -504,8 +504,17 @@ def download(self, globus=True, verbose=False, **kwargs):
...
-            return self
+            dir_length = len(os.listdir(path))
+            if self.dataset.splits:
+                # if metadata indicates splits, check that the directory has as many files as there are splits
+                if(dir_length == len(self.dataset.splits)):
```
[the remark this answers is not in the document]
ascourtas: I would add a print statement informing the user that the file is going to be re-downloaded (in the case that the number of files is not as expected)

foundry/foundry.py
```diff
@@ -529,6 +538,23 @@ def download(self, globus=True, verbose=False, **kwargs):
...
                 }
             xtract_https_download(self, verbose=verbose, **xtract_config)
+
+        # after download check making sure directory exists, contains proper amount of files
+        if os.path.isdir(path):
+            #checking for proper number of files downloaded
+            dir_length = len(os.listdir(path))
+            if self.dataset.splits:
+                if(dir_length != len(self.dataset.splits)):
+                    print("Error: Incorrect number of files in directory")
```
ascourtas, line 548: Excellent, I would just slightly adjust this so you're printing in the exception, like so: https://stackoverflow.com/questions/2052390/manually-raising-throwing-an-exception-in-python

ascourtas (changes requested):

ascourtas (commented):
looks good, well done!

ascourtas (approved):

PR #115, opened 2021-07-12
first commit of no file error

ribhavb:
Fix for clubhouse story: https://app.clubhouse.io/ai-for-materials-and-chemistry/story/493/more-descriptive-error-for-no-file

Also changed the os.path.join code to variable 'path_to_file' where relevant, along with the try excepts for checking validity of path and existence of file.


kjschmidt913 (approved):
Error message looks helpful! And the path_to_file variable seems like a good call.

PR #117, opened 2021-07-13
added resiliency for extra files in download

ribhavb:
in download, instead of checking whether downloaded directory contains exact amount of files as specified in metadata, now checks to make sure it downloads at least as many as specified in metadata for flexibility

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -1,34 +1,36 @@
-from requests.packages.urllib3.exceptions import InsecureRequestWarning
+
+from foundry.xtract_method import *
+import time
+import h5py
+import glob
+import json
+import requests
+import mdf_toolbox
+from json2table import convert
+import pandas as pd
+from datetime import date
+from typing import Any
+import multiprocessing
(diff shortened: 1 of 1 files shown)


blaiszik (changes requested):
I wonder if a more robust approach to this would be to check the existence of the files specified in splits. 

There are two cases that I see:
1) **No splits exist.** Condition: there should be at least one file in the directory
2) **Splits exist.** Condition: There should be a file that corresponds exactly to the file specified in each split.

blaiszik (changes requested):
Could you change the format() functions to Python f' strings?

ascourtas:
@ribhavb @blaiszik is this PR ok now?

blaiszik:
Should this be closed now that we have the other OOP PR? @ribhavb

blaiszik:
I commented in the wrong chain here. Does this still need to be merged since the other download resiliency was merged?
@ribhavb

ribhavb:
Nope! I'll close this now

PR #128, opened 2021-07-25
Fix index issue 

blaiszik:
This fix adds a variable to track the currently specified index e.g., mdf or mdf-test. Without this variable, the context was getting lost and precluding working with production datasets.

Once any comments are addressed and this is merged, let's roll this out to pypi @ascourtas

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -45,6 +45,7 @@ class Foundry(FoundryMetadata):
     dlhub_client: Any
     forge_client: Any
     connect_client: Any
+    index = ""
 
     xtract_tokens: Any
 
@@ -52,6 +53,7 @@ class Foundry(FoundryMetadata):
         self, no_browser=False, no_local_server=False, index="mdf-test", authorizers=None, **data
     ):
         super().__init__(**data)
+        self.index = index
 
         if authorizers:
             auths = authorizers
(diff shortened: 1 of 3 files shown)


foundry/foundry.py
```diff
@@ -165,7 +167,9 @@ def load(self, name, download=True, globus=True, verbose=False, metadata=None, a
 
         del res["projects"][self.config.metadata_key]
 
-        self = Foundry(**res, authorizers=authorizers)
+        # Creating a new Foundry instance is a problematic way to update the metadata,
```
ascourtas, line 170: @blaiszik could you drop a `TODO` in here so we don't lose track of this? I'll make a CH story. Agreed that it's problematic and may lead to other bugs down the road

foundry/foundry.py
```diff
@@ -45,13 +45,15 @@ class Foundry(FoundryMetadata):
...
     connect_client: Any
+    index = ""
 
     xtract_tokens: Any
 
     def __init__(
         self, no_browser=False, no_local_server=False, index="mdf-test", authorizers=None, **data
     ):
         super().__init__(**data)
+        self.index = index
```
ascourtas, line 56: could you just add a quick comment that includes the options available? "mdf-test", "mdf-prod"

PR #134, opened 2021-07-26
Docs

zk794:
deleted stuff from readme


kjschmidt913:
I think we can ditch the "Formal documentation is currently under development." line because we're doing a lot of work on those that by the time people are looking at our documentation, I think it'll be in a pretty good place.

PR #137, opened 2021-08-02
attempted fix to get rid of f=f.load() call in general

ribhavb:
Fix contained in foundry load() function. Not entirely sure if this is the most CORRECT fix since it still involves instantiating an object, but in other attempts the object either wouldn't look quite right or things would go wrong elsewhere 

Also question -- since ideally we won't need to use the f = f.load() call, do we still need the return line at the end, or does it not make a difference? 
(We made sure and tested the code locally that when we removed the return line, everything still worked normally)

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -147,7 +147,7 @@ class Foundry(FoundryMetadata):
         # handle empty dataset name (was returning all the datasets)
         if not name:
             raise ValueError("load: No dataset name is given")
-            
+
         if metadata:
             res = metadata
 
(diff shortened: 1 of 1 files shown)


ascourtas:
> Fix contained in foundry load() function. Not entirely sure if this is the most CORRECT fix since it still involves instantiating an object, but in other attempts the object either wouldn't look quite right or things would go wrong elsewhere
> 
> Also question -- since ideally we won't need to use the f = f.load() call, do we still need the return line at the end, or does it not make a difference?
> (We made sure and tested the code locally that when we removed the return line, everything still worked normally)

this is a great start, but with this story I was looking for more of a comprehensive overhaul of our programming approach with regards to how we treat the `Foundry()` class and instances of it. This would indeed break all of our examples currently, and change how users interact with the library. I will add more context to the story and hopefully that'll help clear things up :) 

you do have the right idea though of making it so load doesn't write over `f`. But you don't want to reinstantiate within the `load()` function, we'd want to just update `self` instead

blaiszik:
Should this be closed now that we have the other OOP PR? @ribhavb

ribhavb:
yup I'll close this.

PR #139, opened 2021-08-03
Dev

blaiszik:


blaiszik:
This commit fixes:
* An error where hdf5 files were being loaded from the wrong path
* An funcX incompatibility issue. funcX bumped to `funcx==0.0.6a5`. We will need to revisit this once Ryan switches DLHub to the new funcX

PR #145, opened 2021-08-06
Bubble vis

zk794:


blaiszik:
Could we add `docs` and `.github` to the ignore list?

ascourtas:
@blaiszik the docs are still visible in the visualization, were you saying you didn't want them there?

I don't think we should add `docs` to the overall gitignore, because that means anytime someone makes a change locally it won't get tracked. 

@ZKatok @zk794

zk794:
ohhh @ascourtas @blaiszik I think i misunderstood, I thought by "ignore" Ben meant the standard gitignore stuff but I suppose he actually meant that he doesn't want the docs shown in the visualization, which makes much more sense. I'll change that now

PR #146, opened 2021-08-09
same pr as download-file-fix, but opened new branch so automated testing works 

ribhavb:


ascourtas:
@ribhavb was this intentionally canceled?

ribhavb:
@ascourtas  ah yup id manually cancelled it because it seemed like it had gotten stuck

PR #147, opened 2021-08-10
tiny change off dev to kick off auto testing

ribhavb:


ascourtas:
whoops didn't realize closing this would remove the test display, but they did indeed all pass.

PR #149, opened 2021-08-11
oop trial number 2 

ribhavb:
oop attempt to fix -- added import for FoundryDatset and created a Fo…undryDataset from the dataset present in the res object in load()


ascourtas:
@ribhavb this looks all set, could you just pull in the most recent changes from `dev` to make sure it still works properly? thank you! sorry this sat for a while

PR #150, opened 2021-08-12
added doi, description, resource type args to publish

ribhavb:
Wasn't sure which new arguments I should put as keyword arguments or just like regular arguments, currently resource_type is a regular argument; description, dataset_doi, and related_dois are keyword args.


blaiszik:
I would remove the resource_type option. In all cases for this publish function, the value should be "Dataset".

PR #151, opened 2021-08-12
added doi test, added test to check hdf5 functionality

ribhavb:
Added 2 tests, edited delete function to also accept dataset source_id as argument


ascourtas:
NOTE: pickle issues resulting in Python 3.7 failure, need to fix first

Aadit-Ambadkar:
This hasn't had any action for a year, should this be closed?

PR #153, opened 2021-08-25
added logic to check if split path begins with /

ribhavb:
Added logic in download to check filepaths for whether they have slashes, adjusts joining and checking accordingly


ascourtas:
@blaiszik unless this is critical path, I'd like to hold off on merging this until we fix the xtract issue so we know the tests are passing. does that sound ok?

ascourtas:
well done!

PR #155, opened 2021-09-29
Release updates for notebooks

kjschmidt913:
Updated notebook examples for upcoming release. Most notable changes:

- made the code for colab vs local environment consistent between notebooks by setting variables for foundry flags at the top
- added specific versions for pandas and tables installations if running in colab
- updated f.load() and f.load_data() lines
- added "index='mdf'" when instantiating foundry (*with one exception, described in the following section)

Things we may want to change before merging this in:

- The fashion and iris datasets weren't in the production datasets. I know we have thought about removing these examples because they aren't materials science examples, so I deleted these 2 notebooks. If we would like to keep them, we should add those datasets to production.
- The bandgaps notebook is currently pulling data from our test datasets. I found one of the datasets in production (foundry_mp_band_gaps_v1.1), but could not find the other (test_double_perovskites_gap_v5.1). I kept this notebook pulling from the test datasets for this reason. If the second dataset is in production but has a different name, just let me know and I'll make the switch.

Documentation has been updated for f.load() and f.load_data() as well.


ascourtas:
@blaiszik what are your thoughts on the test vs prod datasets issues KJ raised?

PR #158, opened 2021-12-16
Aristanascourtas/sc 900/globus false bug

ascourtas:
- fix how foundry writes data files when downloading vis HTTPS and Xtract
- add comments


ascourtas:
@blaiszik here's the PR if you wanna see it!

PR #165, opened 2022-01-14
add joblib to setup.py and requirements.txt, increment to foundry ver…

ascourtas:
…sion 0.0.18


ascourtas:
the branch name here is misleading -- this is just an update for setup.py and requirements.txt to include `joblib`

PR #167, opened 2022-01-18
Nb cleanup

kjschmidt913:
- removed unnecessary package installations
- updated tutorial language to reflect that
- cleared any outputs (installs, pip freeze) that were long and didn't need to be there


kjschmidt913:
i made this branch off of dev, which is where the updated atom finding notebook is. When i updated the atom finding notebook with new language, i got a merge conflict, which is where you see i merged dev into this branch. it was hard to tell in the code of the notebook github gave me to review/resolve. so we just need to double check that in dev after merging in

PR #174, opened 2022-05-04
establish run() function again

ascourtas:
- re-introduce run() function
- does not include DLHub SDK publication methods; thus, will work on Apple Silicon
- requires `dlhub_sdk==0.10.0`
- requires Globus SDK 2, even though `dlhub_sdk==0.10.0` uses Globus SDK 3


ascourtas:
I got an issue where auth was not completing properly in the last run -- I am trying to add `"funcx"` to the list of services to auth, tbd if that fixes things -- previously we invoked funcx directly in auth using `"https://auth.globus.org/scopes/facd7ccc-c5f4-42aa-916b-a0e270e2c2a9/all"`

I wonder if this could be partially due to FuncX service changes recently made.

PR #177, opened 2022-05-11
update reqs for Globus SDK 3, and version to 0.2.0

ascourtas:
- bring package inline with Globus SDK 3
- update requirements accordingly
- update to version `0.2.0` for a minor release


ascourtas:
@blaiszik @kjschmidt913 definitely let me know if ya'll think this should be version `0.2.0` or `0.1.3`

PR #199, opened 2022-06-22
Add keep_hdf5 functionality to Foundry.load_data

Aadit-Ambadkar:


blaiszik:
This looks good. 

A few things:
1) Let's change the `keep_hdf5` flag to `as_hdf5`
2) I think the default for `as_hdf5` should probably be False or None

foundry/foundry.py
```diff
@@ -307,6 +307,8 @@ def load_data(self, source_id=None, globus=True):
         Args:
            inputs (list): List of strings for input columns
            targets (list): List of strings for output columns
+           source_id (string): Relative path to the source file
+           keep_hdf5 (bool): If True and dataset is in hdf5 format, keep data in hdf5 format
```
Aadit-Ambadkar, line 311:
```suggestion
           as_hdf5 (bool): If True and dataset is in hdf5 format, keep data in hdf5 format
```

foundry/foundry.py
```diff
@@ -293,7 +293,7 @@ def run(self, name, inputs, **kwargs):
         """
         return self.dlhub_client.run(name, inputs=inputs, **kwargs)
 
-    def load_data(self, source_id=None, globus=True):
+    def load_data(self, source_id=None, as_hdf5=True):
```
Aadit-Ambadkar, line 296:
```suggestion
    def load_data(self, source_id=None, as_hdf5=False):
```

foundry/foundry.py
```diff
@@ -725,7 +727,7 @@ def get_keys(self, type=None, as_object=False):
             return key_list
 
 
-    def _load_data(self, file=None, source_id=None, globus=True):
+    def _load_data(self, file=None, source_id=None, as_hdf5=True):
```
Aadit-Ambadkar, line 730:
```suggestion
    def _load_data(self, file=None, source_id=None, as_hdf5=False):
```

Issue #29, opened 2020-07-10
PyTorch and Tensorflow dataset compatibility


blaiszik:
* Add capabilities to map Foundry datasets to PyTorch or Tensorflow datasets. 
* Ensure we are capturing enough metadata to make this possible.

### PyTorch
https://pytorch.org/tutorials/beginner/data_loading_tutorial.html

### Tensorflow
https://www.tensorflow.org/datasets

Aadit-Ambadkar:
The problem here (from what I can see) is that Foundry Datasets have little regulation surrounding them. When I run `res = f.load_data()`, I get back a dictionary that is guaranteed to have a 'train' key. But when I access `res['train']`, I can get back a tuple, a dictionary, and possibly other objects. For it to be possible to map a foundry dataset to PyTorch or TF, we would need to standardize the way in which datasets are formated. So, for instance, requiring the structure to be as follows:
```json
{
    "train": {
        "input-type-1": [
            "input1",
            "input2",
            "input3",
            etc
        ],
        "input-type-2": [
            "input1",
            "input2",
            "input3",
            etc
        ],
        etc,
        "target": [
            "target1",
            "target2",
            "target3",
            etc
        ]
    }
}
```

Aadit-Ambadkar:
Ok, so the above problem has (partially) been resolved, so I can guarantee that I'm either getting hdf5 formatted datasets or foundry datasets. I'll be handling hdf5 first, and then foundry. I'll focus on pytorch now (as I am more familiar with that) and once I get pytorch to work, I'll be workint on tenorflow.

Aadit-Ambadkar:
This was merged!

PR #200, opened 2022-06-23
Add Convert to Pytorch Dataset Functionality

Aadit-Ambadkar:
Addresses half of #29.


foundry/foundry.py
```diff
@@ -1,4 +1,6 @@
 
+import enum
+from sklearn import datasets
```
blaiszik, line 3: Is this used?

Aadit-Ambadkar: No, I don't think so. I don't recall adding them either. I've removed them now.

ascourtas:
Hi @Aadit-Ambadkar fantastic work here! Would you mind also adding that test to the GHA (GitHub Actions) test file? That's the one that gets run regularly in continuous integration. 

And also adjust this PR to merge to `dev` instead of `main`, as Ben requested

foundry/foundry.py
```diff
@@ -725,7 +731,7 @@ def get_keys(self, type=None, as_object=False):
             return key_list
 
 
-    def _load_data(self, file=None, source_id=None, globus=True):
+    def _load_data(self, file=None, source_id=None, as_hdf5=False):
```
ascourtas, line 734: did you remove `globus=True` because it was unused? It appears that is a mistake from a past merge -- please add it back in and we will create a separate issue to use `globus` 🙂

requirements.txt
```diff
@@ -6,9 +6,11 @@ six>=1.11.0
 h5py>=2.10.0
 numpy>=1.15.4
 pandas>=0.23.4
+sklearn>=0.0
```
ascourtas, line 9: I believe the proper package is actually `scikit-learn`, based on the `sklearn` package info https://pypi.org/project/sklearn/

ascourtas (changes requested):

foundry/foundry.py
Aadit-Ambadkar: Yes, I did remove it for that reason. However, I don't really see how it could be used, as all the data should already be loaded.

requirements.txt
Aadit-Ambadkar: Done

foundry/foundry.py
Aadit-Ambadkar: Done

foundry/eda.py
```diff
@@ -0,0 +1,27 @@
+import numpy as np
```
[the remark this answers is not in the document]
Aadit-Ambadkar: I've renamed it. Let me know if the naming is more clear now.

PR #201, opened 2022-06-24
To tf dataset

Aadit-Ambadkar:
addresses the rest of #29


foundry/external_data_architectures.py
```diff
@@ -25,3 +26,23 @@ def __getitem__(self, idx):
         
         return item
     
+class FoundryDataset_TF(Sequence):
```
WardLT, line 29: Could you split TF and Pytorch into separate files? That way both do not need to be installed to use one of the loaders.

foundry/external_data_architectures.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 import torch
 from torch.utils.data import Dataset
+from tensorflow.keras.utils import Sequence
 
 class FoundryDataset_Torch(Dataset):
```
WardLT, line 6: Underscores in class names is a little rare in Python. Maybe just call it "FoundryDataset" for Torch and "FoundrySequence" for TF? That way they match the nomenclature for the individual tools

WardLT, line 6: Could you add a short statement about what these classes are for. Just something basic like "Dataset that automatically loads from a Foundry-style resource."

foundry/external_data_architectures.py
```diff
@@ -25,3 +26,23 @@ def __getitem__(self, idx):
         
         return item
     
+class FoundryDataset_TF(Sequence):
+    def __init__(self, inputs, targets):
```
WardLT, line 30: Add in some arguments for these.

foundry/external_data_architectures.py
```diff
@@ -25,3 +26,23 @@ def __getitem__(self, idx):
...
+        self.inputs=inputs
+        self.targets=targets
+
+    def __len__(self):
+        return len(self.inputs[0])
+
+    def __getitem__(self, idx):
+        item = {"input": [], "target": []}
+        
+        for input in self.inputs:
```
WardLT, line 40: Should we care about batching, or can you use this class with a tensorflow dataloader chain?

foundry/external_data_architectures.py
```diff
@@ -25,3 +26,23 @@ def __getitem__(self, idx):
...
+            item["input"].append(np.array(input[idx]))
+        item["input"] = np.array(item["input"])
+        
+        for target in self.targets:
+            item["target"].append(np.array(target[idx]))
+        item["target"] = np.array(item["target"])
+        
+        return item
```
WardLT, line 48: Do we have Flake8 installed for the CI? If so, it should point out style problems like the missing line on some of these files.

foundry/foundry.py
```diff
@@ -857,6 +857,53 @@ def toTorch(self, raw=None, split=None):
             return FoundryDataset_Torch(inputs, targets)
         else:
             raise NotImplementedError
+
+    def toTensorflow(self, raw=None, split=None):
```
WardLT, line 861: Check out the Google Python Style guide: they discourage camelCase and recommend underscore_style. I'd recommend keeping to that here.

foundry/foundry.py
```diff
@@ -857,6 +857,53 @@ def toTorch(self, raw=None, split=None):
...
+        """Convert Foundry Dataset to a Tensorflow Sequence
+
+        Arguments:
+            raw (dict): The output of running ``f.load_data(as_hdf5=False)``
```
WardLT, line 865: I'd explain what the dictionary is without referencing another method.

foundry/foundry.py
```diff
@@ -857,6 +857,53 @@ def toTorch(self, raw=None, split=None):
...
+                    Recommended that this is left as ``None``
+                    **Default:** ``None``
+            split (string): Split to create Tensorflow Sequence on.
```
WardLT, line 868: Consider using type annotations instead of/in addition to typing in the docstring. Might as well start doing that now as it's a feature that is only getting more useful as more Python libraries adopt typing.

foundry/foundry.py
```diff
@@ -857,6 +857,53 @@ def toTorch(self, raw=None, split=None):
...
+            raw = self.load_data(as_hdf5=False)
+        
+        if not split:
+            split = self.dataset.splits[0].type
+
+        if self.dataset.data_type.value == "hdf5":
+            inputs = []
+            targets = []
+            for key in self.dataset.keys:
+                if len(raw[split][key.type][key.key[0]].keys()) != self.dataset.n_items:
```
WardLT, line 884: Complex code (e.g., a 4-nested dictionary access) is going to be hard to maintain later. Make sure to add in some line comments (I like one per ~5 lines of code, fewer if the code is complicated)

requirements.txt
```diff
@@ -14,3 +14,4 @@ mdf-connect-client>=0.4.0
 json2table>=1.1.5
 joblib>=1.1.0
 torch>=1.8.0
+tensorflow>=2.8
```
WardLT, line 17: I strongly vote against adding ML packages as mandatory requirements, as they can be large and difficult to get combined together with other packages. 

We can do optional, or just leave them out and inform users they must install tesnorflow when they want to use specific features.

WardLT (commented):
Really nice work! I couldn't resist reading it over, and had some thoughts while I was at it.

foundry/external_data_architectures.py
Aadit-Ambadkar: There should be ways that you can wrap this class inside another with batching. I don't think it's necessary for us to include batching functionality.

foundry/foundry.py
Aadit-Ambadkar: We should do this in another PR for all of the functions in the Foundry Package

foundry/external_data_architectures.py
Aadit-Ambadkar: not sure what you mean by this

foundry/foundry.py
Aadit-Ambadkar: I just removed that argument all together as it isn't useful, and can cause code to break

foundry/external_data_architectures.py
WardLT: Whoops, that comment did make no sense. 

Add in some documentation for these arguments

ascourtas:
@Aadit-Ambadkar per @WardLT's feedback, we're requesting adding documentation under the Advanced Usage section in GitBook for how to use TF and PyTorch Datasets with Foundry! ✨ These docs will ensure both other developers and our users will know how to use these features effectively. 

https://app.gitbook.com/o/-MW5XeB7GD8rLHu7Mw6m/s/-MW6A6E0L1zra4m3voL_-38885591/advanced-usage/pytorch-datasets 

Please also add the notebook you have under the `/examples` folder in this repo

foundry/foundry.py
WardLT: Agreed, but let's start now too

foundry/tensorflow_dataset_wrapper.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+from tensorflow.keras.utils import Sequence
+
+class FoundryDatasetAsTensorflowSequence(Sequence):
+    """Foundry Dataset Converted to Tensorflow Format"""
+
+    def __init__(self, inputs, targets):
```
WardLT, line 7: You should document what the inputs to these classes are.

foundry/tensorflow_dataset_wrapper.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
```
WardLT, line 1: Instead of having two Python files with similar, lengthy names in the same module, let's create a new module. Perhaps something like `foundry.loaders` and have a `tf` and `pytorch` submodule in each. The file layout would look like

```
foundry\
  __init__.py
[ ... ]
 loaders\
    __init__.py
    tf.py
    [ ... ]
```

foundry/tensorflow_dataset_wrapper.py
```diff
@@ -0,0 +1,25 @@
...
+        
+        for input in self.inputs:
+            item["input"].append(np.array(input[idx]))
+        item["input"] = np.array(item["input"])
+        
+        for target in self.targets:
+            item["target"].append(np.array(target[idx]))
+        item["target"] = np.array(item["target"])
+        
+        return item
```
WardLT, line 25: There is a newline missing at the end of the file, which is a requirement in Python's styleguide.

GH Actions hasn't been raising these style errors (see #208 ) but we should fix them anyway.

foundry/foundry.py
```diff
@@ -854,9 +852,52 @@ def toTorch(self, raw=None, split=None):
...
+        """Convert Foundry Dataset to a Tensorflow Sequence
+
+        Arguments:
+            split (string): Split to create Tensorflow Sequence on.
+                    **Default:** ``None``
+
+        Returns: (FoundryDatasetAsTensorflowSequence) Tensorflow Sequence of all the data from the specified split
+
+        """
+        raw = self.load_data(as_hdf5=False)
```
WardLT, line 869: Move the import statement for the TF data loader into this function so that Foundry won't load Tensorflow until needed. Imports not at the top is a non-standard practice, but useful when adding functions which rely on optional requirements (e.g., Pandas does this: https://github.com/pandas-dev/pandas/blob/v1.4.3/pandas/core/generic.py#L2190-L2353)

foundry/foundry.py
```diff
@@ -854,9 +852,52 @@ def toTorch(self, raw=None, split=None):
...
+            targets = []
+
+            for index, arr in enumerate([inputs, targets]):
+                df = raw[split][index]
+                for key in df.keys():
+                    arr.append(df[key].values)
+
+            return FoundryDatasetAsTensorflowSequence(inputs, targets)
+        else:
+            raise NotImplementedError
```
WardLT, line 899: 👏🏼 Good to have a "default" action that alerts the user to problems with an exception.

WardLT (changes requested):
Nice improvements! 
- Still need improvements on documentation - both module level (that'll get picked up by autodoc tools) and line comments within the source code
- Better handling of optional requirements (TF/pytorch are large packages that should not be required)
- Restructuring to avoid lengthy module/class names, and promote better organization of new features.
-

Aadit-Ambadkar:
> Nice improvements!
> 
> * Still need improvements on documentation - both module level (that'll get picked up by autodoc tools) and line comments within the source code
> * Better handling of optional requirements (TF/pytorch are large packages that should not be required)
> * Restructuring to avoid lengthy module/class names, and promote better organization of new features.

Documentation in GitBook is done (I think), I'll get to work on the other parts.

foundry/tensorflow_dataset_wrapper.py
Aadit-Ambadkar: really anything. all foundry datasets have data separated into inputs and targets. so inputs can really be anything, and that has to be the case since it would be ridiculously hard to accommodate for every dataset otherwise.

foundry/foundry.py
Aadit-Ambadkar: I can't because the Data Loader extends a TF Sequence.

foundry/foundry.py
Aadit-Ambadkar: I'm not quite sure what this means.

WardLT: That the if/elif statement throws an error if the code ends up matching a condition you didn't expect when implementing it. The "default" behavior is the term for such a condition in C/C++/Java.

Aadit-Ambadkar: I'm still not entirely sure what you mean. Like a try catch?

foundry/foundry.py
```diff
@@ -28,9 +28,8 @@
     FoundryDataset
 )
 
-from foundry.external_data_architectures import (
-    FoundryDataset_Torch
-)
+from foundry.loaders import *
+
```
WardLT, line 32: Avoid imports of "*" unless you've got great reason for it. Delaying loading of libraries helps improve responsiveness of Python code, and using `import *` loads everything before you can do any work.

foundry/foundry.py
```diff
@@ -812,21 +810,18 @@ def _load_data(self, file=None, source_id=None, globus=True, as_hdf5=False):
...
             split (string): Split to create PyTorch Dataset on.
                     **Default:** ``None``
 
-        Returns: (FoundryDataset_Torch) PyTorch Dataset of all the data from the specified split
+        Returns: (TorchDataset) PyTorch Dataset of all the data from the specified split
 
         """
-        if not raw:
-            raw = self.load_data(as_hdf5=False)
+        from foundry.loaders.torch_wrapper import TorchDataset
```
WardLT, line 823: The `import *` is definitely not needed if you import it later, like this.

WardLT (changes requested):
This is _so so_ close to merging. There's an import statement you missed removing and the code still lacks line comments.

foundry/foundry.py
```diff
@@ -854,9 +849,54 @@ def toTorch(self, raw=None, split=None):
...
+                    continue
+
+                val = np.array([raw[split][key.type][key.key[0]][k] for k in raw[split][key.type][key.key[0]].keys()])
+                if key.type == 'input':
+                    inputs.append(val)
+                else:
+                    targets.append(val)
+
+            return TensorflowSequence(inputs, targets)
+        elif self.dataset.data_type.value == "tabular":
```
WardLT, line 887: Another quick thought: What do you think of moving this code, which is specific to the TFSequence to a @classmethod of TensorflowSequence rather than inside of this function?

Aadit-Ambadkar:
> I just read the code again to understand it better: my last suggestion about a classmethod was wrong. I see now that the code here is used across multiple "to_dataset" functions which means it should be moved to a new "make inputs/outputs" functions.
> 
> Also, do you see my complaints about line comments? I keep adding one each time I review, but I'm not sure you're getting the message.

I'm not able to see your complaints about line comments.

foundry/foundry.py
Aadit-Ambadkar: This was marked as resolved so I didn't check it... oops.

foundry/loaders/tf_wrapper.py
```diff
@@ -0,0 +1,26 @@
...
+        self.inputs=inputs
+        self.targets=targets
+
+    def __len__(self):
+        return len(self.inputs[0])
+
+    def __getitem__(self, idx):
+        item = {"input": [], "target": []}
+        
+        for input in self.inputs:
```
WardLT, line 17: I was poking through this code again, and am a little concerned about this part. 

We're combining all of our inputs into a single array, which will only work with the following line that combines them into a single array if they are the same shape. 

I can see this working if your different inputs are, for example, different color channels from the same image, but not if one is an image and the other is a 1D array associated with it.

I'm not yet sure what real-world datasets we want to use with this, so I don't have recommendations for what the behavior should be yet.

Could you write a line comment about what this loop is supposed to do to make it easier to tune later?

ascourtas:
@Aadit-Ambadkar bump on this

foundry/loaders/tf_wrapper.py
Aadit-Ambadkar: Since it is not a numpy array, they don't need to all have the same shape.

Aadit-Ambadkar:
I think this is good to merge, I'll see what the conflicts are though

WardLT:
Do you mean 'ready to merge' as 'it has no conflicts?' I still see some issues around duplicated code requiring introducing a new method.

Aadit-Ambadkar:
@WardLT could you let me know where those are?

foundry/foundry.py
Aadit-Ambadkar: I think this is done?

foundry/foundry.py
```diff
@@ -804,23 +801,8 @@ def _load_data(self, file=None, source_id=None, globus=True, as_hdf5=False):
...
-            split (string): Split to create PyTorch Dataset on.
-                    **Default:** ``None``
-
-        Returns: (FoundryDataset_Torch) PyTorch Dataset of all the data from the specified split
-
-        """
-        if not raw:
-            raw = self.load_data(as_hdf5=False)
+    def _get_inputs_targets(self, split: str = None):
+        raw = self.load_data(as_hdf5=False)
```
WardLT, line 805: Document this function:
- What does this do? 
- What does the "split" parameter mean"
- What does it return?

tests/test_foundry.py
```diff
@@ -116,11 +105,11 @@
     'n_items': 1000
 }
 
-# Globus endpoint for '_iris_dev'
-test_data_source = "https://app.globus.org/file-manager?origin_id=e38ee745-6d04-11e5-ba46-22000b92c6ec&origin_path=%2Ffoundry-test%2Firis-dev%2F"
+# Globus endpoint for '_iris_dev' for test publication
```
WardLT, line 108: @ascourtas , it seems like we're adding a new file here. Is this PR out of sync with the current testing strategy?

foundry/loaders/tf_wrapper.py
WardLT, line 17: Please document what these for loops are doing. I'm still dubious about returning a single array that contains multiple inputs is correct here, so let's at least explain.

The ideal solution would be to test with datasets that represent what we have in foundry. @ascourtas , does anything come to mind?

WardLT (changes requested):
Code looks ready to accept, but the documentation is lacking and I'm concerned that we're about to see a merge problem with the tests (@ascourtas , could you check on that?).

tests/test_foundry.py
Aadit-Ambadkar: hmm, I merged with def, that could be why

foundry/foundry.py
Aadit-Ambadkar: Ok, fixed this.

Aadit-Ambadkar:
@WardLT do you know why the checks are failing?

WardLT (approved):
LGTM. I'm not yet convinced this'll work for all datasets, but we should definitely get it out into the wild to figure that out. 

Last concern is one for @ascourtas / @blaiszik : Which of the two test files that this PR has is correct?

WardLT:
The testing might be broken because the underlying testing CI scripts have changed. Let's merge and then fix if there's a problem, as I'm now worried about this drifting too far from dev

ascourtas:
> LGTM. I'm not yet convinced this'll work for all datasets, but we should definitely get it out into the wild to figure that out.
> 
> Last concern is one for @ascourtas / @blaiszik : Which of the two test files that this PR has is correct?

thanks for flagging this Logan, I'll take a look now

.gitignore
```diff
@@ -2,3 +2,4 @@
 *.DS_STORE
 *.pyc
 *.idea
+*/foundry_ml.egg-info/*
```
ascourtas, line 5: please add a newline at the end of this file

tests/test_foundry.py
ascourtas: @WardLT @Aadit-Ambadkar yes this is out of date -- currently in `dev` we have one test file (as it should be) with logic specifying which functions should be run in GHA and which are local-only. https://github.com/MLMI2-CSSI/foundry/blob/dev/tests/test_foundry.py 

my suggestion would be to pull the most recent changes to `dev` into this branch before merging into `dev`

foundry/loaders/tf_wrapper.py
```diff
@@ -24,4 +23,4 @@ def __getitem__(self, idx):
         item["target"] = np.array(item["target"])
         
         return item
-    
+
```
ascourtas, line 26: all of these files appear to be missing a proper newline at the end? please amend

ascourtas (changes requested):
I'd also suggest revisiting the inline comments -- there's some explanation here of what is happening, but not a lot of "why". The reader should ideally be able to read only the comments and know exactly what the code is doing.

ascourtas:
> LGTM. I'm not yet convinced this'll work for all datasets, but we should definitely get it out into the wild to figure that out.
> 
> Last concern is one for @ascourtas / @blaiszik : Which of the two test files that this PR has is correct?

test_foundry.py is the correct one; answered here https://github.com/MLMI2-CSSI/foundry/pull/201#discussion_r944708831

foundry/loaders/tf_wrapper.py
ascourtas: yes the iris data used in the test files (should be in the right format), or the Atom Position Finding dataset (HDF5) plus the Zeolite dataset (tabular) would give one a comprehensive test

ascourtas (approved):
Issues with conflicting test files and merge inconsistencies have been handled; the only remaining edits are mostly comments, which are documented [here](https://github.com/MLMI2-CSSI/foundry/issues/247) -- otherwise, this is good to merge

Issue #211, opened 2022-07-01
handle multiple matches in `load()`


kjschmidt913:
In `Foundry.load()`, currently we assume that Forge will only return one matching dataset/model in Foundry (or at least, we only take the first one returned).

Amend this to handle the case where multiple matches may be returned.

the ideal behavior here would be to throw a warning to the user if multiple datasets are returned

See TODO in below code snippet

     def load(self, name, version="1.1", provider="MDF", download=True, globus=True, verbose=False, **kwargs,):
        """Load the metadata for a Foundry dataset into the client
        Args:
            name (str): Name of the foundry dataset
            download (bool): If True, download the data associated with the package (default is True)

        Keyword Args:
            interval (int): How often to poll Globus to check if transfers are complete

        Returns
        -------
            self
        """
        res = []
        # MDF specific logic
        # Handle DOI inputs
        if name.startswith("10.") and provider == "MDF":
            print("Loading by DOI")
            res = (
                self.forge_client.match_dois(name)
                .match_resource_types("dataset")
                .match_field("mdf.organizations", "foundry")
                .search()
            )

        # Handle MDF source_ids
        else:
            print("Loading by source_id")
            res = (
                self.forge_client.match_field("mdf.organizations", "foundry")
                .match_resource_types("dataset")
                .match_field("mdf.source_id", name)
                .search()
            )

        if not res:
            return self
        else:
            # res is list of all matches in Forge
            # TODO: handle if there are multiple matches
            res = res[0]

            # map result to FoundryMetadata Pydantic structure (see models.py)
            res["dataset"] = res["projects"]["foundry"]
            res["dataset"]["type"] = res["dataset"]["package_type"]
            del res["projects"]["foundry"]

            # TODO: reassign values to self in a safer way
            self = Foundry(**res)

        if download is True:  # Add check for package existence
            self.download(
                interval=kwargs.get("interval", 10), globus=globus, verbose=verbose
            )

        return self

Aadit-Ambadkar:
This was done 😳

PR #227, opened 2022-07-07
Fix .load function

Aadit-Ambadkar:
Fix .load function according to #211

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -35,7 +35,9 @@ from requests.packages.urllib3.exceptions import InsecureRequestWarning
 import logging
 import os
 import shutil
+import warnings
 logging.disable(logging.INFO)
+
 
 
 class Foundry(FoundryMetadata):
@@ -169,9 +171,6 @@ class Foundry(FoundryMetadata):
         if metadata:
             res = metadata
 
-        if metadata:
-            res = metadata
-
         # MDF specific logic
         if is_doi(name) and not metadata:
             res = self.forge_client.match_resource_types("dataset")
@@ -185,7 +184,9 @@ class Foundry(FoundryMetadata):
 
         # unpack res, handle if empty
         try:
-            # if search returns multiple results, this automatically uses first result
+            # if search returns multiple results, this automatically uses first result, while warning the user
+            if len(res) > 1:
+                warnings.warn("Multiple datasets found for the given search query")
             res = res[0]
         except IndexError as e:
             raise Exception("load: No metadata found for given dataset") from e
@@ -199,7 +200,7 @@ class Foundry(FoundryMetadata):
 
         # TODO: Creating a new Foundry instance is a problematic way to update the metadata,
         # we should find a way to abstract this.
-
+        
         fdataset = FoundryDataset(**res['dataset'])
         self.dc = res['dc']
         self.mdf = res['mdf']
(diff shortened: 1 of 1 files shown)


foundry/foundry.py
```diff
@@ -185,7 +184,9 @@ def load(self, name, download=True, globus=True, verbose=False, metadata=None, a
 
         # unpack res, handle if empty
         try:
-            # if search returns multiple results, this automatically uses first result
+            # if search returns multiple results, this automatically uses first result, while warning the user
+            if len(res) > 1:
+                warnings.warn("Multiple datasets found for the given search query")
             res = res[0]
         except IndexError as e:
```
WardLT, line 191: An issue that looks like it predates your PR: Using try/except is an excessive route for testing whether a list is empty. 

Since you are already adding a "if len" test, could you convert this to statement to have 

```python
if len(res) == 0:
   raise Exception("load: No metadata found for dataset \"{name\"") 
```

Note that I also put the name of the dataset into the exception.

foundry/foundry.py
```diff
@@ -199,11 +200,10 @@ def load(self, name, download=True, globus=True, verbose=False, metadata=None, a
...
-
-        fdataset = FoundryDataset(**res['dataset'])
+        
         self.dc = res['dc']
         self.mdf = res['mdf']
-        self.dataset = fdataset
+        self.dataset = FoundryDataset(**res['dataset'])
 
 
         if download is True:  # Add check for package existence
```
WardLT, line 209: Sorry man, another thing caught my attention? Why `is True`? Why not just `if dataset`?

WardLT (commented):
This looks great to me, but I realized that you probably meant for @ascourtas to review this.

My only complaints are on code that was here before this PR. Any chance you'd not mind fixing those while you're at it?

foundry/foundry.py
Aadit-Ambadkar: oh I have no clue, I'll fix that though.

Aadit-Ambadkar:
> This looks great to me, but I realized that you probably meant for @ascourtas to review this.
> 
> My only complaints are on code that was here before this PR. Any chance you'd not mind fixing those while you're at it?

I also meant for you to do so (because you're just so dang good at reviewing PRs).

foundry/foundry.py
ascourtas: @Aadit-Ambadkar once this fix is added we can merge! Everything else looks good.

Aadit-Ambadkar:
@ascourtas

foundry/foundry.py
```diff
@@ -35,9 +35,11 @@
 import logging
 import os
 import shutil
+import warnings
 logging.disable(logging.INFO)
 
 
+
```
ascourtas, line 42: this is really small, but the whitespace should be adjusted such that there are 2 lines between `import warnings` and `logging.disable()`, and then only 2 lines after `logging.disable()`, not 3

foundry/foundry.py
```diff
@@ -184,11 +183,13 @@ def load(self, name, download=True, globus=True, verbose=False, metadata=None, a
             res = res.match_field("mdf.source_id", name).search()
 
         # unpack res, handle if empty
-        try:
-            # if search returns multiple results, this automatically uses first result
-            res = res[0]
-        except IndexError as e:
-            raise Exception("load: No metadata found for given dataset") from e
+        if len(res) == 0:
+            raise Exception("load: No metadata found for given dataset")
```
ascourtas, line 187: add name of dataset please

ascourtas (changes requested):

ascourtas (approved):
much better, great work!

Issue #221, opened 2022-07-05
Printing to screen for non-critical errors is unnecessary 


WardLT:
There are a few instances where we print directly to stdout from code. That's rarely used in Python scripts as it is difficult to control whether `print` goes to screen or not. People are stuck with it unless they mess with `sys.stdout` (not recommended). 

Consider switching these to logging statements instead.

Example: https://github.com/MLMI2-CSSI/foundry/blob/7b9427df2b12ea4f2880e5e550d5880cde37768f/foundry/foundry.py#L764

blaiszik:
Is this completed?

WardLT:
Yes, I don't see any more print statements.

Issue #224, opened 2022-07-05
Data importing logic is overly complicated


WardLT:
We use nested try-except blocks to attempt to determine a data type currently, and there are a few alternatives to this which should be considered:

1. _Provide the list of possible functions as a list._ You can still use try/except to evaluate each one, but it will avoid having to nest them inside each other. Just break the loop if one passes and continue if not.
1. _Don't rely on automatically guessing each time you load the dataset._ Get the format once on publication then read it from the Foundry record. We had code to do this back in a super-early version of DLHub SDK: https://github.com/DLHub-Argonne/dlhub_sdk/blob/v0.4.0/dlhub_sdk/models/datasets.py#L70

https://github.com/MLMI2-CSSI/foundry/blob/7b9427df2b12ea4f2880e5e550d5880cde37768f/foundry/foundry.py#L759

blaiszik:
#271 addresses Part 1.

PR #228, opened 2022-07-07
 Fix Logging and Remove unnecessary code

Aadit-Ambadkar:
#221


Aadit-Ambadkar:
@WardLT do you know why the tests are failing? It says that a file is not present but the tests do preload a file.

foundry/foundry.py
```diff
@@ -672,30 +657,6 @@ def build(self, spec, globus=False, interval=3, file=False):
                 key_list = key_list + k
             return key_list
 
-        print("Building Data Package")
```
ascourtas, line 675: @blaiszik what is the purpose of the `build()` function? We don't appear to use it anywhere. 

Aadit, I am wary of cutting out this chunk of the function without more context (also, can you provide your reasoning for axing this chunk?)

foundry/foundry.py
```diff
@@ -761,21 +722,21 @@ def _load_data(self, file=None, source_id=None, as_hdf5=False):
                     path_to_file
                 )
             except Exception as e:
-                print(f"Reading {file} as JSON failed: {e} \n", "Now attempting to read as JSONL")
+                logging.log(logging.WARNING, f"Reading {file} as JSON failed: {e} \n", "Now attempting to read as JSONL")
```
ascourtas, line 725: small note -- I think it would be better to say `"Cannot read {file} as JSON: {e} \n", "Now attempting to read as JSONL"` since seeing the word "failed" is a little misleading (and concerning!)

foundry/foundry.py
```diff
@@ -747,7 +708,7 @@ def _load_data(self, file=None, source_id=None, as_hdf5=False):
             try:
                 path_to_file = os.path.join(path, file)
             except Exception as e:
-                print(f"Unable to find path to file for download: {e}")
+                logging.log(logging.FATAL, f"Unable to find path to file for download: {e}")  # is this even necessary?
```
ascourtas, line 711: Hm it looks like there is some conflicting comments/intentions here. `os.path.join()` will rarely error out afaik. 

My suggestion would be to remove this try/except block, unless @kjschmidt913 has any reasons we shouldn't. The important thing is to check that the file exists, which is done in the next block

foundry/foundry.py
ascourtas: ditto the following, similar except blocks

foundry/foundry.py
```diff
@@ -596,15 +581,15 @@ def download(self, globus=True, verbose=False, **kwargs):
                         missing_files.append(split.path)
                 #if number of missing files is greater than zero, redownload with informative message
                 if len(missing_files) > 0:
-                    print(f"Dataset will be redownloaded, following files are missing: {missing_files}")
+                    logging.log(25, f"Dataset will be redownloaded, following files are missing: {missing_files}")
```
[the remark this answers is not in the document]
ascourtas: ditto the following

ascourtas (changes requested):

foundry/foundry.py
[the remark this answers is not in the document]
Aadit-Ambadkar: The number is because we disabled Info Level logging but this is not a Warning (so I put something in between)

foundry/foundry.py
Aadit-Ambadkar: Oh that chunk of context can't be run ever. The if else statment has return clauses, so the function always returns in that segment. The chunk is never run.

foundry/foundry.py
WardLT: ☝🏼 Ari's on point. `join` will  fail if file or path is `None` but that's a roundabout why of testing what we're looking for. We should directly test those conditions rather than rely on a secondary mechanism.

Plus, it is better to contain the error message in the exception (e.g., `raise ValueError('Unable to find path...')`)

Putting it together:

```python

if path is None or file is None:
    raise ValueError(........)
path_to_file = os.path.join(path, file)
```

foundry/foundry.py
WardLT: I think WARNING is overblown. Warnings imply someone should do something about how they are using the code, and them providing JSON is no worse or better than using CSV. Use INFO or, even, DEBUG instead.

WardLT: FWIW, I think this whole block needs some TLC #224

WardLT (changes requested):
I second @ascourtas 's comments.

ascourtas:
@Aadit-Ambadkar just wanted to bump on this

foundry/foundry.py
ascourtas: Agreed on both accounts! INFO or DEBUG would be much better (probably INFO in this case)

foundry/foundry.py
WardLT: @Aadit-Ambadkar , you still haven't addressed this comment about the use of a try/except block when an if statement is clearer.

I don't think we need to rely on the FATAL to notify people when also throwing an exception.

foundry/foundry.py
```diff
@@ -761,21 +721,21 @@ def _load_data(self, file=None, source_id=None, as_hdf5=False):
...
-                print(f"Reading {file} as JSON failed: {e} \n", "Now attempting to read as JSONL")
+                logging.log(logging.INFO, f"Reading {file} as JSON failed: {e} \n", "Now attempting to read as JSONL")
                 try:
                     # Try to read individual lines instead
                     self.dataset.dataframe = pd.read_json(
                         path_to_file, lines=True
                     )
                 except Exception as f:
-                    print(f"Reading {file} as JSONL failed: {f} \n", "Now attempting to read as CSV")
+                    logging.log(logging.INFO, f"Reading {file} as JSONL failed: {f} \n", "Now attempting to read as CSV")
```
WardLT, line 731: See above comment about using a module-level logger.

WardLT (changes requested):

foundry/foundry.py
Aadit-Ambadkar: Sorry, not quite sure what you mean by "module" (is that function or class?)

foundry/foundry.py
Aadit-Ambadkar: I think what was intended was os.path.exists(os.path.join(...,...))

foundry/foundry.py
```diff
@@ -1,38 +1,26 @@
 
+from multiprocessing.sharedctypes import Value
```
WardLT, line 2: Is this import doing anything? If not, delete it. If so, we need to talk why we're touching this lower-level part of Python

WardLT (changes requested):
There seems to be a wayward import. Beyond that, this looks ready to merge.

WardLT:
@ascourtas , should we merge this into `dev` and not `main`?

Aadit-Ambadkar:
> @ascourtas , should we merge this into `dev` and not `main`?

oh shoot yeah this was supposed to go to main.

foundry/foundry.py
ascourtas: @WardLT I assume you still want this deleted, yes? @Aadit-Ambadkar only part of it appears to be deleted based on the diff

ascourtas (changes requested):

foundry/foundry.py
ascourtas: @Aadit-Ambadkar

ascourtas (changes requested):

foundry/foundry.py
ascourtas: @Aadit-Ambadkar this still needs attention

foundry/foundry.py
ascourtas: @Aadit-Ambadkar could you please address this block as well.

ascourtas (changes requested):

foundry/foundry.py
Aadit-Ambadkar: I only deleted the part that was dead. The rest is still "alive." I only deleted part of the function which could never be accessed (there was a return statement above that)

Aadit-Ambadkar:
@ascourtas I think you are looking at outdated code? Could you check again?

foundry/foundry.py
ascourtas: @Aadit-Ambadkar I believe we determined with @blaiszik and @WardLT that the entire function is deprecated, please delete the whole thing unless Ben and Logan disagree

foundry/foundry.py
```diff
@@ -759,21 +720,21 @@ def _load_data(self, file=None, source_id=None, globus=True, as_hdf5=False):
                     path_to_file
                 )
             except Exception as e:
-                print(f"Reading {file} as JSON failed: {e} \n", "Now attempting to read as JSONL")
+                logger.info(f"Reading {file} as JSON failed: {e} \n", "Now attempting to read as JSONL")
```
ascourtas, line 723: @Aadit-Ambadkar this still needs to be amended, please see other comments

ascourtas (changes requested):

foundry/foundry.py
ascourtas: removed with newest commits

ascourtas (approved):
removed deprecated `build()` function, amended log statements for reading tabular data, simplified and made DRY-er some of the path-handling logic for loading datasets

Should be good to merge now once we fix the CI testing issues. Tests pass locally.

Aadit-Ambadkar:
@ascourtas tests working now.

Aadit-Ambadkar:
oops ignore that

PR #229, opened 2022-07-14
fix unused imports, code style, and update syntax

isaac-darling:
DO-NOT-MERGE label is added because I cannot confirm the accuracy of my changes with 100% certainty.


WardLT:
Do you want to try adding in a GitHub action that uses flake8 to check for new errors now that you've finished fixing the errors?

foundry/xtract_method.py
```diff
@@ -135,7 +135,7 @@ def download_file(file):
         return {file["path"] + " status": True}
 
     results = Parallel(n_jobs=num_cores)(
-        delayed(download_file)(file) for file in file_ls
```
WardLT, line 138: This seems wrong: the `delayed` wrapper in joblib is what makes it execute the code in parallel. As written, this will execute the function and then submit it to be executed rather than run things asynchronously.

isaac-darling: I'm not sure, all I changed there was I switched to the decorator syntactic sugar so the line is easier to read

WardLT: 🤦I looked for and, somehow, missed that you did use the decorator.

Nothing to see here! Carry on!

WardLT:
I approve and vote to squash-merge since we have a lot of minor commits

PR #241, opened 2022-08-06
Extract xtract

blaiszik:


foundry/foundry.py
```diff
@@ -820,3 +827,53 @@ def is_doi(string: str):
...
+                for item in res["DATA"]
+                if item["type"] == "dir"
+            )
+        for item in res["DATA"]:
+            if item["type"]=='file':
+                item["name"] = path_prefix + item["name"]
+                item["path"] = abs_path.replace('/~/','/')
+                yield item
+
+def recursive_ls(tc, ep, path, max_depth=3):
```
WardLT, line 853: Add some docstrings to these

foundry/foundry.py
```diff
@@ -820,3 +827,53 @@ def is_doi(string: str):
...
+    queue = deque()
+    queue.append((path, "", 0))
+    yield from _get_files(tc, ep, queue, max_depth)
+
+def download_file(item, https_config):
```
WardLT, line 858: Could you add a docstring and/or type hints? E.g., what content does "item" have and where does it usually come from? 

Additionally, the indentation is odd here. Does your editor have an autopep you could run on this?

WardLT (changes requested):
Just documentation and code style complaints. 

We're also missing tests to make sure that this functions correctly. Can you do something where we xfer to a globus test endpoint?

ascourtas:
closed since Ben already merged this into `main`

Issue #237, opened 2022-07-29
add query functionality to Foundry for searching within `f.list()`


ascourtas:
`f.list()` is getting long, we should add simple functionality for users to query it

blaiszik:
@blaiszik added this functionality, and it has been merged in.

PR #252, opened 2022-09-01
 Add filters for f.list

Aadit-Ambadkar:
Addresses #237


foundry_ml.egg-info/PKG-INFO
```diff
@@ -0,0 +1,56 @@
+Metadata-Version: 2.1
```
ascourtas, line 1: Make sure you remove this file! should not be committed

foundry_ml.egg-info/SOURCES.txt
```diff
@@ -0,0 +1,17 @@
+LICENSE
```
ascourtas, line 1: please remove

foundry_ml.egg-info/dependency_links.txt
```diff
@@ -0,0 +1 @@
+
```
ascourtas, line 1: please remove this file

foundry_ml.egg-info/requires.txt
```diff
@@ -0,0 +1,10 @@
+mdf_forge>=0.8.0
```
ascourtas, line 1: I see what's going on here -- please remove the entire `foundry_ml.egg-info` folder :)

foundry/foundry.py
```diff
@@ -208,12 +208,15 @@ def load(self, name, download=True, globus=True, verbose=False, metadata=None, a
...
+            name_filter (callable): A function or lambda which takes in the name of the dataset and returns a boolean; if None, not used
+            year_filter (callable): A function or lambda which takes in the year of publication of the dataset and returns a boolean; if None, not used
 
         Returns
         -------
-            (pandas.DataFrame): DataFrame with summary list of Foundry data packages including name, title, publication year, and DOI
+            (pandas.DataFrame): DataFrame with summary list of Foundry data packages including name, title, publication year, and DOI, given the filters.
         """
         res = (
             self.forge_client.match_field(
```
ascourtas, line 222: Hi @Aadit-Ambadkar ! I think instead of filtering after the fetch, we want to use Forge and Globus SDK methods to search the database appropriately. Otherwise, when we eventually have 1000s of datasets, it'll be inefficient to postprocess (and also, it would be nice for the search parameters to be a little more flexible). @blaiszik has already started working on this issue, so I think it would be best to let him finish it.

ascourtas (commented):
Hi @Aadit-Ambadkar , I'm so happy to see you adding things! I think this one requires some more in-depth Globus SDK finagling that I'll leave to @blaiszik , but I love that you're picking up issues!

foundry/foundry.py
Aadit-Ambadkar: Pandas is reliably (like ridiculously) fast, potentially faster than using Forge / Globus SDK methods, but I think it makes sense.

PR #253, opened 2022-09-01
Open in Colab Buttons

Aadit-Ambadkar:
Addresses #251


ascourtas (changes requested):
These look great, well done Aadit! My only requested change would be to please move the button up to be directly under the title of each notebook, and to be left-aligned. 

Feel free to break the title out into a separate cell to make this happen

Aadit-Ambadkar:
@ascourtas they are left aligned...

I'll move them all up to be just below the title soon

PR #256, opened 2022-09-01
Add search

blaiszik:
Adding new Foundry `search()` function. Search allows users to supply a query, and specify a max number of results to find closer matching datasets than `list()` which returns every Foundry dataset.

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -197,18 +197,23 @@ class Foundry(FoundryMetadata):
 
         return self
 
-    def list(self):
-        """List available Foundry data packages
+    def search(self, q=None, limit=None):
+        """Search available Foundry data packages
+        q (str): query string to match
+        limit (int): maximum number of results to return
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


ascourtas:
@blaiszik please address flake8 concerns :)

foundry/foundry.py
```diff
@@ -199,18 +199,23 @@ def load(self, name, download=True, globus=True, verbose=False, metadata=None, a
 
         return self
 
-    def list(self):
-        """List available Foundry data packages
+    def search(self, q=None, limit=None):
+        """Search available Foundry data packages
```
ascourtas, line 203: iirc I believe we decided not to use the "package" language -- please change this to "datasets" or whatever you see fit

ascourtas (changes requested):
tiny comment suggestion -- besides that and the flake8 changes, should be good to go!

ascourtas (approved):
looks good!

PR #273, opened 2022-10-21
Paralellize HTTPS downloads. Remove joblib and six requirements

blaiszik:


foundry/foundry.py
```diff
@@ -485,9 +486,9 @@ def download(self, globus=True, verbose=False, **kwargs):
             task_list = list(recursive_ls(self.transfer_client,
                                           https_config['source_ep_id'],
                                           https_config['folder_to_crawl']))
-            # TODO Add parallel
             for task in task_list:
-                download_file(task, https_config)
+                with ThreadPoolExecutor() as executor:
+                    executor.submit(download_file, task, https_config)
```
WardLT, line 491: A quick note on this one: moving the for loop inside the with statement would allow you to limit the number of parallel downloads and threads. Large datasets with many files could be a problem

foundry/foundry.py
```diff
@@ -485,9 +486,9 @@ def download(self, globus=True, verbose=False, **kwargs):
             task_list = list(recursive_ls(self.transfer_client,
```
WardLT, line 486: If you want to be fancy, don't use a list first. That way you can download while finding the files (I'm assuming recursive ls is a generator)

WardLT (commented):
Noticed a few things

PR #274, opened 2022-10-21
Get citation function

blaiszik:
`Foundry.get_citation()` returns a BibTeX string of the dataset citation


ascourtas (approved):
looks good to me!

PR #276, opened 2022-10-24
Improve the HTTPS downlaods

WardLT:
A few changes, but the big one is letting the downloads actually happen in parallel.

Others include fixing docstrings, removing dead code, consolidating test requirements to a file, and adding a progressbar.


WardLT:
Superceded by #277

Issue #193, opened 2022-06-15
enable dataset publishers to publish using HTTPS


ascourtas:
Publishing a dataset using HTTPS is now possible that we're using Eagle, which has Globus Connect Server 5. We should expose this functionality to our users such that they can publish a dataset using HTTPS and don't need to upload the data to a Globus Endpoint, share the source_url, etc.

ascourtas:
this is higher priority now that Tyler's Xtract service will be deprecated

ascourtas:
Steps outlined in the Miro Board here: 

https://miro.com/app/board/o9J_lfSY3Kg=/?moveToWidget=3458764532712986360&cot=14

ascourtas:
Much success in building basic functionality in working sessions on 9/8. Remaining tasks include:

High level
- [ ] add tmp dirs to Eagle, then Transfer to UIUC using regular MDF methods
- [ ] integrate helper functions into Foundry's `publish()`
- [ ] add tests

Granular
- [ ] add capability to upload more than one file
- [ ] add deletion of ACL rule
- [ ] add `prepare` parameter to `put`
- [ ] set try/except clause to catch if ACL rule has been set before, and if dir has been made before
- [ ] add functionality to create tmp dirs 
- [ ] parallelize to send multiple files at once?

ascourtas:
pseudocode on `publish()` setup

```
publish(add new arg https=True, local_path='')
    if https:
        link = https_upload(path)
        data_source = [link]
    
    ...
    ...
    ...
    
    if https:
        remove_acl()
```

ascourtas:
Mostly there as of today, end-to-end functionality should be good. Just need to add some more bells and whistles, documentation, and tests.

ascourtas:
Actually need to remove usage of `prepare` and test recursive directory-adding instead; prepare does not work as intended (after playing around with it, it appears to make directories out of the filenames, regardless of the order `prepare` is specified in)

ascourtas:
Still struggled with `?prepare` -- it would result in no files being uploaded even though I got a 200 from the PUT request. Going to axe this for now

ascourtas:
Currently in PR https://github.com/MLMI2-CSSI/foundry/pull/281

kjschmidt913:
note: doesn't currently work with client credentials

ascourtas:
done! just need to update docs and cut release

Issue #279, opened 2022-11-03
Parallelize HTTPS upload 


ascourtas:
_As a Dataset Publisher, I want to be able to upload my data quickly and easily_

# Assumptions

1. HTTPS upload functionality works properly already
2. Parallelizing does not impact the written directory structure
3. Parallelizing does not impact the publisher's reception of error messages

# Acceptance Criteria
1. Given I have a dataset to upload, I can do so in rapid time (ideally less than 1 min), even for datasets that are many GB in size 
2. Assumptions 2 and 3 are still met 

# Tasks

- [ ] Parallelize existing HTTPS upload code in `publish()`

kjschmidt913:
since https upload will be tackled during publishing refactor, closing this

PR #281, opened 2022-11-04
Add https upload

ascourtas:
Allows users to publish a dataset to Foundry using HTTPS instead of providing a data source URL. Underneath, this uploads the data to the NCSA endpoint using an HTTPS PUT request, and then Globus Transfers the data from there to the MDF endpoint. The data are written to `/tmp` on the NCSA endpoint, organized by publication UUID

- sets HTTPS upload to True by default
- does not include parallelization (https://github.com/MLMI2-CSSI/foundry/issues/279)
- known issue where creating an ACL rule can fail for a user with `Not authorized for the endpoint globuspublish#mdf-publications (82f1b5c6-6e9b-11e5-ba47-22000b92c6ec)` but the user can still write to the endpoint 

NOTE: tests still need to be added, should not be merged until they are 

Fixes https://github.com/MLMI2-CSSI/foundry/issues/193


foundry/foundry.py
```diff
@@ -39,7 +44,9 @@ class Foundry(FoundryMetadata):
     forge_client: Any
     connect_client: Any
     transfer_client: Any
+    auth_client: Any
```
WardLT, line 47: Defining the fields up here is unnecessary unless we add documentation to them.

foundry/foundry.py
```diff
@@ -365,12 +392,201 @@ def publish(self, foundry_metadata, data_source, title, authors, update=False,
...
+                files={
+                    'file': (filename, f, 'application/octet-stream')
+                }
+            )
+        if reply.status_code != 200:
+            raise IOError(f"Error on HTTPS PUT, got response {reply.status_code}: {reply.text}")
+        # Return the response
+        return reply
+
+    def make_globus_link(self, endpoint_id, path):
```
WardLT, line 575: How do you feel about moving these functions out into an "upload" submodule? I'm hoping to cut down on the number of functions of the base object to make it simpler to understand how to use

foundry/foundry.py
```diff
@@ -365,12 +392,201 @@ def publish(self, foundry_metadata, data_source, title, authors, update=False,
...
+                    raise IOError(f"Error while creating child directory {dest_path}: {e.message}") from e
+            # get local path to file to upload
+            for filename in files:
+                filepath = os.path.join(root, filename)
+                # upload file to destination path on endpoint
+                result = self._upload_file(filepath, https_base_url, dest_path, endpoint_id)
+                results.append(result)
+        return results
+
+    def _upload_file(self, filepath, https_base_url, dest_path, endpoint_id):
```
WardLT, line 538: Prime candidate for moving into a submodule: Not a function that people should use from the FoundryClient, but one that might get used elsewhere.

foundry/foundry.py
```diff
@@ -320,20 +332,31 @@ def get_citation(self) -> str:
...
+            publication_year=None, test=False, **kwargs,):
         """Submit a dataset for publication
         Args:
             foundry_metadata (dict): Dict of metadata describing data package
-            data_source (string): Url for Globus endpoint
             title (string): Title of data package
             authors (list): List of data package author names e.g., Jack Black
                 or Nunez, Victoria
+            https_data_path (str): Path to the local dataset to publish to Foundry via HTTPS. Creates an HTTPS PUT
+                request to upload the data specified to a Globus endpoint (default is NCSA endpoint) before it is
```
WardLT, line 345: Do we need to cite the default is NCSA here? I'm looking for ways to shorten the docstring.

foundry/foundry.py
```diff
@@ -365,12 +392,201 @@ def publish(self, foundry_metadata, data_source, title, authors, update=False,
...
+            (list): A list of all of the HTTPS PUT request results (dicts) from the uploads
+        """
+        results = []
+        # initialize destination path as the parent destination path
+        dest_path = parent_dest_path
+
+        # walk through each child directory in the designated local data folder
+        for root, directories, files in os.walk(local_data_path):
+            # update destination path if we have walked into a child directory
+            if root is not local_data_path:
```
WardLT, line 520: If these are strings, it should be "!=" and not "is not." (you have have two different strings objects with the same value)

WardLT: `os.walk` doesn't guarantee it will pass the same object back as input.

tests/test_foundry.py
```diff
@@ -242,6 +242,11 @@ def test_globus_dataframe_load():
     _delete_test_data(f)
 
 
+def test_https_upload():
+    f = Foundry(authorizers=auths)
```
WardLT, line 246: Should we have a test here?

foundry/foundry.py
```diff
@@ -365,12 +392,201 @@ def publish(self, foundry_metadata, data_source, title, authors, update=False,
...
+        else:
+            res = None
+
+        # if uploaded by HTTPS, delete ACL rule after dataset submission is complete
+        if https_data_path and rule_id:
+            self.transfer_client.delete_endpoint_acl_rule(endpoint_id, rule_id)
+
         return res
 
+    def _create_dest_folder(self, endpoint_id, parent_dir="/tmp"):
```
WardLT, line 425: Do we want to use type annotations through our library? I'm in favor as it will improve IDE experience

foundry/foundry.py
```diff
@@ -365,12 +392,201 @@ def publish(self, foundry_metadata, data_source, title, authors, update=False,
         self.connect_client.add_organization(self.config.organization)
         self.connect_client.set_project_block(
             self.config.metadata_key, foundry_metadata)
-        self.connect_client.add_data_source(data_source)
+
+        # upload via HTTPS if specified
```
WardLT, line 396: If this operation is just a wrapper over a method in the connect client, then why aren't we putting the main logic into Connect? I'd imagine people want this capability for MDF in general?

WardLT (commented):
I think the code's great, but question why it is part of the FoundryClient:

1. This operation is just a shell over ConnectClient. Why aren't we adding the core logic to Connect so that it is available to all MDF users?
2. Many of the private methods for the object are useful in their own right. Why not break them off into a separate module? That would also simplify testing

ascourtas:
> I think the code's great, but question why it is part of the FoundryClient:
> 
> 1. This operation is just a shell over ConnectClient. Why aren't we adding the core logic to Connect so that it is available to all MDF users?
> 2. Many of the private methods for the object are useful in their own right. Why not break them off into a separate module? That would also simplify testing

Good question! @blaiszik shall I move this over to MDF instead?

foundry/foundry.py
ascourtas: I feel great about that! There's certainly a lot of helper functions needed for this

foundry/foundry.py
ascourtas: good point, waiting on Ben's feedback

foundry/foundry.py
ascourtas: Hm I guess I don't see how removing that parenthetical would significantly reduce the size of the docstring overall? But I do think it adds quick clarity to what this function is doing, such that a developer could just read the docstring and not the code to know where the data are being written to

foundry/foundry.py
ascourtas: Good point, I am still stuck in my old ways and forget about type annotations. I will add them ✨

foundry/foundry.py
ascourtas: whoops good catch, thank you Logan!

tests/test_foundry.py
ascourtas: yes, perhaps this wasn't easily visible but I mentioned in the PR comment that I don't want to merge this until I finish the test -- I was just trying to get the code out there before our Friday sprint planning and my brief vacation, so I could get feedback sooner.

foundry/foundry.py
WardLT: Looking again, I've got a different impression and am feeling kinda surprised I was advocating for less docs. 

Maybe I'd put the explanation of the two different modes for supplying the data earlier in the docstring so that people can understand how to get the desired behavior (e.g., upload via HTTPS) without having to read each argument.

ascourtas: that sounds like a great addition! can do

tests/test_foundry.py
WardLT: Makes sense. Shows that I didn't read well enough :)

foundry/foundry.py
ascourtas: Ben responded on Slack: yes to moving this over to MDF! I will open a PR in https://github.com/materials-data-facility/connect_client and when that's up, I will close this one

ascourtas:
TODOs post review:

- [ ] add type annotations in function definitions
- [ ] move helper functions to a separate submodule 
- [ ] change path comparison operator to "!=" in `os.walk()` block
- [ ] in the docstring for `publish()`,  put the explanation of the two different modes for supplying the data earlier in the docstring so that people can understand how to get the desired behavior (e.g., upload via HTTPS) without having to read each argument
- [ ] finish tests

ascourtas:
@blaiszik @WardLT I have addressed most of the requested changes, but I got a little hung up on design choices for "move helper functions to a separate submodule" -- I have a test that tests just `_upload_to_endpoint()` so we can test the majority of the functionality without actually publishing to MDF. 

However, when I break these out into a submodule, they are no longer a part of the `Foundry` class and thus do not have the appropriate authorizers, etc. 

Do ya'll have any suggestions on the best course of action?

ascourtas:
Hm, it's now also failing with an error I didn't see when I ran things locally: 

```
>           raise IOError(f"Error from Globus API while creating destination folder: {e.message}") from e
E           OSError: Error from Globus API while creating destination folder: Denied by endpoint, Error (mkdir)
E           Endpoint: globuspublish#mdf-publications (82f1b5c6-6e9b-11e5-ba47-22000b92c6ec)
E           Server: data.materialsdatafacility.org:443
E           Command: MKD /tmp/f2673963-6d6f-4ad4-b331-9b26c3dde1dd
E           Message: Fatal FTP Response
E           ---
E           Details: 500 Path not allowed.\r\n
```

WardLT:
Wrt the separate module, could you have the functions take a Transfer client as input?

ascourtas:
initial WIP code for the submodule breakout can be found here: https://github.com/MLMI2-CSSI/foundry/tree/add_https_upload_submodule

isaac-darling:
Mostly done here but a few questions:
1. I correctly guessed that the error before my last commit was a one-time thing, (no code change in the recent commit) but that possibly leaves unanswered questions.
2. I added docstrings for the `auths` param on autopilot here since I lack some context on the workings of Globus auth, so they may be vague at best and incorrect at worst.
3. There may be a better variable to pass around than `auths`, but I unsure of what that might be.

isaac-darling:
Update on number 1 from above: is it possible that two tests get the current time simultaneously causing this error? If so, it seems that should be changed.

foundry/https_upload.py
```diff
@@ -0,0 +1,238 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def upload_to_endpoint(auths: dict, local_data_path: str, endpoint_id: str = "82f1b5c6-6e9b-11e5-ba47-22000b92c6ec",
+                       dest_parent: str = None, dest_child: str = None) -> Tuple[str, str]:
+    """Upload local data to a Globus endpoint using HTTPS PUT requests. Data can be a folder or an individual file.
+        Note that the ACL rule created in this method must later be deleted after the dataset is submitted to MDF.
+    Args:
+        auths (dict): the required authorizers
```
WardLT, line 31: Would it be possible to pass the TransferClient and AuthClient as inputs? That way, users don't need to know how the dictionary must be structures (e.g.,, 'transfer' -> TransferClient)

ascourtas: yes I think that makes sense! good catch, I didn't realize how much unnecessary unpacking happens downstream

ascourtas: Although actually @WardLT , I don't know if it would make that much of a difference -- currently we need the AuthClient,  the TransferClient, and the authorizer for the NCSA scope (see this line). So to pass all 3 of those, I think I'd want to bundle them in an object anyways, which means we'll still need to reference by key. 

What do you think? 

If we do keep the object passing, I could write more in the docstrings so that the user knows what keys map to what. And perhaps make it a smaller object than full `auths`

WardLT: I think this is more of an aesthetic decision then. I prefer having many arguments for a command so that the docstring is simpler, an IDE can help you catch when an argument is wrong, and you avoid having to write your own error messages for badly-formatted dictionaries. (e.g., `f(auth={'trnfser': client})` would not be detectable by an IDE, but `f(trnsfer=client)` would and would also throw an error in Python) 

That said, I'd back your play if you choose to keep the "put it all in a dictionary" approach

WardLT: Perhaps a `dataclass` or `NamedTuple` that holds the auth clients rather than a dictionary would be a middle ground: you could keep all of the authorizers together, but you would put up more guiderails of what the "authorizer collection" can contain.

ascourtas: these are great points! I'm going to think about it a bit and do some googling. I really like everything you raised here. 

My goal was readability and ease of invoking a function. 

I like the idea of a NamedTuple

WardLT: Sounds good. Are you thinking still that the dictionary is the right object?

ascourtas: I actually liked the NamedTuple best! felt like the best balance of user-friendly and lightweight (dataclass is thorough but felt like overkill, dict felt less user friendly since you need to know keys).

ascourtas:
all right @WardLT , how about now? I did not add asserts 1. Because my hunch was right, it is indeed [ill-advised](https://stackoverflow.com/questions/944592/best-practice-for-using-assert) per the Internet, and 2. There is no way to check the `endpoint_id` used in the scope for an `AuthClient` after it's been instantiated, so it was a bit of a mess. 

Instead, I switched to passing the Dict `endpoint_auth_clients` around, such that a user would never have an `endpoint_id` and their desired `AuthClient` misaligned. 

Hopefully that's suitable!

foundry/foundry.py
```diff
@@ -351,6 +376,10 @@ def publish(self, foundry_metadata, data_source, title, authors, update=False,
             of dataset. Contains `source_id`, which can be used to check the
             status of the submission
         """
+        if (https_data_path and globus_data_source) or \
+                (https_data_path is None and globus_data_source is None):
+            raise ValueError("Must assign either `https_data_path` or `globus_data_source`")
```
WardLT, line 381: Minor note: you can use XOR here (`http_data_path ^ globus_data_source`)

foundry/https_upload.py
```diff
@@ -0,0 +1,248 @@
+"""
+https_upload.py
```
WardLT, line 2: A filename in the docstring for a file is superfluous. I'd recommend deleting.

foundry/https_upload.py
```diff
@@ -0,0 +1,248 @@
...
+
+Private utility methods to upload files and/or folders to Globus using HTTPS instead of Globus Transfer.
+
+Authors:
+    Aristana Scourtas
+    Isaac Darling
+
+Last modified 12/21/22 by Isaac Darling
```
WardLT, line 10: The author list and modification dates are covered by git, and likely to fall out-of-date. Let's delete them before they are just noise.

foundry/https_upload.py
```diff
@@ -0,0 +1,248 @@
...
+            or altering data on a specific endpoint).
+        local_data_path (str): The path to the local data to upload. Can be relative or absolute.
+        https_base_url (str): The URL for a given Globus endpoint.
+        parent_dest_path (str): The path to the parent folder to be written to on the given endpoint. The contents
+            of "local_data_path" will be written here, including subdirectories.
+        endpoint_id (str): The UUID designating the exact Globus endpoint. Can be obtained via the Globus Web UI or
+            the SDK. This must be the same endpoint pointed to by the https_base_url.
+    Returns
+    -------
+        (list): A list of all the HTTPS PUT request results (dicts) from the uploads
```
WardLT, line 174: Is this return statement still right? I saw the function it calls returns a `Response`

foundry/https_download.py
```diff
@@ -1,3 +1,10 @@
+"""
+https_download.py
```
WardLT, line 2: As above. Don't need to say the filename in the docstring. It will surely go out of date.

foundry/https_download.py
```diff
@@ -1,3 +1,10 @@
...
+
+Methods to download files from a Globus endpoint using Xtract (HTTPS). Now deprecated as of Aug 2022
```
WardLT, line 4: Do we need to say this replaces Xtract? I doubt it will help many people understand the code.

foundry/https_upload.py
```diff
@@ -0,0 +1,248 @@
...
+        endpoint_auth_clients (Dict[str, AuthClient]): Dict of `endpoint_id` : `AuthClient` keypairs. AuthClients used
+            for Globus Auth functionality within endpoint-specific scopes using Globus Connect Server (ie accessing
+            or altering data on a specific endpoint).
+        local_data_path (str): The path to the local data to upload. Can be relative or absolute.
+        https_base_url (str): The URL for a given Globus endpoint.
+        parent_dest_path (str): The path to the parent folder to be written to on the given endpoint. The contents
+            of "local_data_path" will be written here, including subdirectories.
+        endpoint_id (str): The UUID designating the exact Globus endpoint. Can be obtained via the Globus Web UI or
+            the SDK. This must be the same endpoint pointed to by the https_base_url.
+    Returns
```
WardLT, line 172: All of the doc strings mix NumPy and Google Styles. ("Args:" is Google, "Return\n-----" is NumPy). I see that the whole library does this, so fix it here if you want.

WardLT (approved):
Looks good to me. My only complaints are docstring-related. I don't think they're anything I need to review again, so merge when you are ready :highfive:

ascourtas:
I think in the next iteration of this (when it moves to MDF), I would like to make an Uploader class so the auths are handled at init and don't require passing around -- feels smelly. 

And then also, we need to make a separate story for cleaning up the Foundry auths

foundry/https_download.py
ascourtas: Actually I think what this is saying is that this file is largely deprecated, because most of it uses xtract. I'd need to look closely at it to see if any of it is still used (I did not write this code, I just moved it out of Foundry a while back). 

I think this is a separate ticket.

foundry/https_upload.py
ascourtas: thanks for the catch! eagle eyes

foundry/https_upload.py
ascourtas: I am glad you brought this up -- this is something I have been doing since I was taught to do so in my early 20s, but it occurs to me now that this pattern probably came from the olden days when there weren't version control systems and it was file chaos. Thanks for challenging this Logan, I haven't thought about it!

foundry/https_download.py
WardLT: This does seem like a separate ticket. I'll resolve

WardLT:
Docstring changes look good to me. Do you want to "squash and merge" or shall I?

ascourtas:
> Docstring changes look good to me. Do you want to "squash and merge" or shall I?

thanks Logan! I've got this

PR #283, opened 2022-11-17
Set Header Images to an absolute URL via raw.githubusercontent

cyschneck:
The README uses relative URL links, so the header image is not visible outside of the github repo (like on pypi)

image


blaiszik:
Thank you for the fix @cyschneck!

PR #320, opened 2023-03-21
Update paper.md

ianfoster:
Note change in second paragraph of summary: "with an initial focus on materials science and chemistry"


ascourtas:
I assume Ben already included any necessary changes in the overleaf? closing until further notice

PR #336, opened 2023-04-27
Update publishing notebook and minor bugfixes

ascourtas:
- update notebook content to be in line with current documentation
- update notebook to include HTTPS upload as the primary way of uploading
- update notebook images for getting Globus link 
- fix error outputted to user on ACL rule creation by removing ACL rule creation (turns out it was unnecessary)
- fix XOR bug


ascourtas:
looks like a lot of commits because we had previously squashed and merged (which we should do again here), but it's not that many -- the diff is valid. most changes are in the jupyter notebook

blaiszik (changes requested):
Minor change requested: In the notebook cell `When will my data package be available in Foundry?` mention that the dataset needs to be approved by a human which can take up to a day or longer around weekends and holidays.

ascourtas:
got all clear from Ben on Slack!

Issue #189, opened 2022-06-08
Set up Sphinx Autodocs


kjschmidt913:
Set up [Sphinx](https://www.sphinx-doc.org/en/master/index.html) with our GitBook docs. Publish from the `docs` GitBook branch.

blue442:
Just for clarification, is the intent here to publish the .md content from the docs folder via sphinx, or are we looking to document the foundry code itself? Also are we looking to deploy it via github pages, or readthedocs, or something else?

blaiszik:
This is mostly completed minus a permissions issue with the bot pushing back to `main`. Should we unprotect the branch, or are there other ways around this?



![Image](https://github.com/MLMI2-CSSI/foundry/assets/4753182/e5dba8e0-40de-46dc-b682-5518efc99335)

blaiszik:
I've verified that this works without branch protection. We can create a new issue to deal with the auth aspects if needed. Do we suspect we actually need branch protection? We already limit who can push via access control.

PR #342, opened 2023-05-19
automating api documentation using github action

blue442:
This PR addresses #189 by creating a github action that automatically builds API documentation from docstrings. The action is set up to be triggered by a successful PR merge, and then generates the .md files into the `docs/` folder of the PR base branch. The action employs the [lazydocs](https://github.com/ml-tooling/lazydocs) package as after extensive exploration, it was the simplest and cleanest solution that I found to generate vanilla .md files. Note that there was no clear way to generate API documentation from the docstrings that integrated with gitbook - as such, the .md files created are not currently integrated into the gitbook. This documentation would also benefit from additional development to refine the docstrings.

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -1,0 +1,42 @@
+name: build api documentation
+
+on:
+  pull_request:
+    types:
+      - closed
+  push:
+    branches:
+      - "*"
+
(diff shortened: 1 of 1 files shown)


blaiszik (commented):
Reviewed and approved

Issue #316, opened 2023-03-02
Add capability to download specific splits


blaiszik:
Given that splits enumerate files to download, we should implement a way to download specific splits. This has come up in 2 contexts (NISt Foam DB, and Ryan Jacobs work)

Some splits may be quite large, and users would only want certain splits.

Example
```

tr = f.load_data(split="train")

```

blaiszik:
Currently all splits are downloaded here: https://github.com/MLMI2-CSSI/foundry/blob/317d8c1b67eaf927a5718e55adc5d5eb2397b724/foundry/foundry.py#LL295C14-L295C14

A user should be able to specify a list of splits (strings) to download in the `load_data()` function.

PR #344, opened 2023-05-25
Split specification

blue442:
Addressing #316 - added the ability to specify a list of splits by label for loading when using the `load_data()` function. Added some ValueError exceptions if there are splits that are requested but not in the dataset, as well as if splits are requested but no splits are present.

Note that testing around the lack of splits in the dataset is currently skipped, as there aren't any obvious datasets that don't have splits that can be used for testing, and from discussions it sounds like that will be protected against in the near future.

Also updated the docstring for `load_data()` to comply with google docstring format and reflect that it returns a `dict`.

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -266,7 +266,7 @@ class Foundry(FoundryMetadata):
             self.dlhub_client.fx_endpoint = funcx_endpoint
         return self.dlhub_client.run(name, inputs=inputs, **kwargs)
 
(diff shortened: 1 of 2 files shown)


blaiszik (changes requested):
Minor change. What do you think about changing `splits_to_load` arg to simply `splits`. I feel like that might read easier in the docstring. Also, the doctoring has it already documented as `splits`

blaiszik (approved):

PR #347, opened 2023-05-31
adding/updating example notebooks of 4 microscopy-related models

wdwzyyg:
Four example notebooks of running models have been tested with dlhub-sdk-2.0.3 foundry-ml-0.6.2 funcx-1.0.13. Now upload them to the main repository so that they can be easily accessed for Paul's coming talk in June.


ascourtas (changes requested):
Hi @wdwzyyg , would you mind removing the `/.ipynb_checkpoints` folders here? they add unnecessary bloat to the repo. Thank you!

PR #356, opened 2023-05-31
Merege Split specification from dev into main (#344)

blaiszik:
* adding ability to specify splits for loading

* refining test

* Update splits_to_load --> splits

---------


ascourtas (approved):
nice, looks good!

Issue #323, opened 2023-04-05
change load() and load_data() to be more descriptive


kjschmidt913:
These names can be confusing for users. Could this be load_metadata and get_data? Or something like that to make them more descriptive?

blaiszik:
What if we moved the args for `load()` into the Foundry constructor instead? and had somethings like this.

```python
f = Foundry("{{DOI or name}}", other args...)
f.get_data()
```

`load()` could then become basically a private method and we'd simplify the naming AND the usage

PR #358, opened 2023-06-01
Load on init

blue442:
Addressing #323 - moved the `Foundry.load()` function entirely into `Foundry.__init__()` and converted it to a private function `Foundry._load()`. Most of the arguments that were passed to the `Foundry.load()` call have been moved to arguments in the `Foundry.__init__()` function but for two exceptions:

1) I needed to get rid of the duplicate keyward argument references in the two functions; the `**data` reference in `Foundry.__init__()` appeared to be more critical, as the `**kwargs` was only referenced in the `Foundry.load()` function definition to extract an interval value with an assumed default of 10 (`kwargs.get('interval', 10)`). As such, I eliminated the `**kwargs` argument from `Foundry.load()` and just replaced with an argument called `interval` which is given a default value of 10 in the call to `Foundry.__init__()`.

2) The `authorizers` argument common to both `Foundry.load()` and `Foundry.__init__()` were assumed to be the same; as such, the pre-existing argument submitted to `Foundry.__init__()` is just passed through to `Foundry._load()`

As a sidenote, the `name` argument was made optional, and the call to `_load()` dependent on the presence of a `name` argument. This was done to facilitate the publishing situations described in the tests where a Foundry instance is instantiated with no call to `Foundry.load()`.

## Testing
Some of the testing was effectively reduced, as the intermediate steps are lost between instantiation and load. Ended up removing some `_delete_test_data(f)` references that occurred mid-test (previously between `Foundry()` and `f.load()`). Not clear to me why the data was deleted inbetween the calls to `Foundry.__init__()` and 'Foundry.load()`; I could be missing something?

Two tests are currently failing; `foundry_tests.test_publish_with_https()` and `foundry_tests.test_publish_with_globus()`, but both are due to 403 responses returned from the submission attempt - I assume these are because I lack permissions vs broken logic.

Note that if this PR passes, we'll have to update the example notebooks (and documentation?) to reflect the new usage.


blaiszik (commented):
Looks great! Could you re-issue this PR against `main`? We have moved to a `main` -> `production` instead of `dev` -> `main` flow.

blaiszik (approved):

Issue #218, opened 2022-07-01
add better error handling if dataset metadata are incorrect


kjschmidt913:
Add descriptive errors when users submit datasets with incorrect metadata

kjschmidt913:
Bumping this since it was brought up as an issue for Wisconsin in December

blaiszik:
This seems like a good path to me. Could you give us a first pass example?

ascourtas:
merged!

PR #377, opened 2023-06-26
Metadata error handling

blue442:
Addressing issue #218 - the work on this branch focuses on ensuring that the metadata is validated as soon as a call to `publish()` is made. This was accomplished by changing the base class that the `Foundry` object extended (from `FoundryMetaData` to `FoundryBase`) - the `FoundryBase` model definition contains all the functionality that was previously in `FoundryMetaData`, but now the `FoundryMetaData` model definition has been repurposed to validate the metadata of the foundry object.

Validation of the metadata is performed when calling `foundry.publish_dataset()` by instantiating an instance of a `models.FoundryDataSet` object using the `foundry_metadata: Dict[str, Any]` argument passed in to the call to `foundry.publish_dataset()`. The `FoundryDataSet` (and supporting models) are defined in the models.py file. The instantiated `FoundryDataSet` is never actually used other than to validate the dictionary passed in to the call to `publish_dataset()`.

Some exception handling and error reporting were included to facilitate communication with the user about what (if any) issues arose during validation.

Additional discussion can be viewed in the github issue (#218).


ascourtas (approved):
Hi Steve, this code is excellent. A very neat solution for an annoying problem, and I learned more about Pydantic by reviewing! The test is succinct and effective. I have no suggestions, it's perfect imo ✨ Approved!

Issue #379, opened 2023-06-30
Break `load()` out of Foundry class `__init__` to improve workflow


ascourtas:
_As a User, I want to be able to know what a foundry function does without reading docs. I want a clear and intuitive workflow for using datasets_

There was some miscommunication in #323 (PR #358 ) about how best to disambiguate `load()` from `load_data()`. The solution developed entails privatizing `load()` by loading upon instantiation of the `Foundry` class, but this disrupts user workflows for loading multiple datasets, using `list()` or `search()`, etc. 

The new solution is to break `load()` out of `__init__`, but instead disambiguate by renaming and potentially further breaking up the functions. For example, one proposed solution is:
- `load()` --> `download_dataset()` and `view_metadata()`, OR `view_dataset()` where `download` is an optional boolean parameter
- `load_data()` --> `create_dataframe()`

Or renaming/refactoring solutions are possible, and require reading through the existing `load()` and `load_data()` implementations to decide on next steps

# Assumptions:
1. You are familiar with the contents of `load()` and `load_data()` and understand what they currently do
2. The implementation in #358 is currently in the `main` branch and needs to change per this ticket
3. We cannot cut a release unless we either change that or do some git wrangling
4. You have looked over the [Example Workflows](https://miro.com/app/board/uXjVPrIdUhE=/?moveToWidget=3458764558322356672&cot=14) defined in our Miro brainstorm

# Acceptance Criteria
1. Given I am a new user of Foundry, I can easily look at the list of functions in my IDE and know exactly what to do to download a dataset, preview the metadata, and use it in a dataframe format
2. As a User, I can accomplish any of the [User Workflows](https://miro.com/app/board/uXjVPrIdUhE=/?moveToWidget=3458764558322356672&cot=14) defined without running into friction based on the way downloading, loading, and previewing datasets is defined

```[tasklist]
### Tasks
- [x ] Review User Workflows (linked above) and discuss user needs and next steps with team
- [ ] Decide on naming solution for `load()` vs `load_data()` and any necessary refactor (see suggested example above)
- [ ] Implement changes 
```

kjschmidt913:
+ update documentation + example notebooks to reflect new syntax

kjschmidt913:
Is this sort of getting the boot since we are rethinking class structures?

ascourtas:
Blocked on class restructure #395

PR #384, opened 2023-07-06
Reviving load

blue442:
Addressing #379, implemented changes to separate the loading of the metadata and dataset from the instantiation of a foundry object in order to facilitate browsing/searching of datasets _prior_ to downloading. In addition, I refactored the downloading of metadata and data to utilize clearer verbiage when calling the functions.

## Specific changes
- Made `load()` a public function again
- Removed the `metadata` arg from `load()`
- Removed `if metadata: res = metadata` from `load()` method
- if `is_doi(name) and not metadata:` -> `if is_doi(name):`
- Remove call to `load()` from the `__init__()` function
- Changed `f.load()` -> `f.fetch_data()`
- Changed `download=True` to `metadata_only=False`

## New usage example snippet

```
import foundry
f = foundry.Foundry()
f.search()
f.fetch_data('foundry_assorted_computational_band_gaps_v1.1', metadata_only=True)
f.download_full_dataset()
data_dictionary = f.load_data()
```

## Documentation notes:
- `load_data()` loads the data into memory, while `download_data()` downloads it to disk, correct? If so, the docstring should probably reflect this.
- Moved 'todos' from code into docstrings (google-style)

## TODO:
- Need to update gitbook
- Need to update notebooks
- From `__init__()` - remove `metadata (dict): **For debug purposes.** A search result analog to prepopulate metadata.`
- Not sure about "# TODO: Creating a new Foundry instance is a problematic way to update the metadata, we should find a way to abstract this." (I don't understand this comment)
- Why do we only return the first of multiple search results? We're already getting all of the results and just throwing all but the first away - should we handle this differently?
- Calling `load()` (now `fetch_data()`)- does a user need to pass in the 'source_id' generated by `list()` as the name argument, or can they pass in the entry under 'name'? ('name' doesn't seem to work)


ascourtas (changes requested):
Hi Steve! I was discussing this solution with @blaiszik, and he pointed out that there wouldn't be a case where the user would want to download the dataset but not load the metadata (but there are cases where a user _would_ want to load the metadata, but not download the dataset). He suggested the following refactor which I think is very smooth:

> "I think we should just make the metadata loading occur without user intervention. It’s a very cheap operation, and I don’t see really any use to have the Foundry object without the metadata"

```
from foundry-ml import Foundry
f = Foundry("$DOI$") # this gets the metadata, and never gets the data
f.get_data() # this downloads the data
X,y = f.load_data() #this loads the data into RAM
```

I think `get_data()` could also be called `download_dataset()`, depending on what we think is clearest. What are your thoughts?

blue442:
Updated based on feedback to reflect three possible use cases:

- initialize foundry object w/o name (for publishing purposes)
- initialize foundry object w/ name, downloads metadata AND dataset
- initialize foundry object w/ name and pass `metadata_only=True` so dataset is not downloaded

Also changed function names to `download_dataset` and `download_metadata` for clarity.

Upgraded `python-setup` in github action to v4 (was causing an error).

foundry/foundry.py
```diff
@@ -155,38 +154,24 @@ def __init__(
...
 
-        if name is not None:
-            self._load(name=name,
-                       download=download,
-                       globus=globus,
-                       verbose=verbose,
-                       metadata=metadata,
-                       authorizers=authorizers,
-                       interval=interval)
+        # handle empty dataset name (was returning all the datasets)
```
ascourtas, line 157: I'm not sure I understand this comment, could you give me a little more context?

foundry/models.py
```diff
@@ -155,7 +155,7 @@ def _repr_html_(self):
 class FoundryBase(BaseModel):
     dc: Optional[Dict] = {}  # pydantic Datacite?
     mdf: Optional[Dict] = {}
-    dataset: FoundryDataset = {}
+    metadata: FoundryMetadata = {}
```
ascourtas, line 158: could you clarify for me the logic of removing the `dataset` field altogether?

ascourtas (commented):
Just need a little clarity on things before I can approve!

foundry/foundry.py
blue442: This comment got kicked around and ended up in a weird place - I put it back in the `download_metadata` with some logic to ensure it wasn't called without a name argument, which appears to download all the datas.

foundry/foundry.py
```diff
@@ -155,38 +154,24 @@ def __init__(
...
+        if name:
+            self.download_metadata(name)
+
+        if name and not metadata_only:
```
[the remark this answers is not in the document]
blue442: Done.

foundry/foundry.py
```diff
@@ -51,7 +50,7 @@ class Foundry(FoundryBase):
 
     def __init__(
             self, name=None, no_browser=False, no_local_server=False, index="mdf", authorizers=None,
-            download=True, globus=True, verbose=False, metadata=None, interval=10,
+            verbose=False, metadata=None, interval=10, metadata_only=False, globus=True,
```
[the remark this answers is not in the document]
blue442: Things got moved around so much the docstring got lost in the shuffle - fixed!

foundry/models.py
ascourtas: resolved in slack convo, removed deprecated fields

ascourtas (changes requested):
Thanks for the changes Steve, almost there! Last thing for this PR: can you please remove the unnecessary added cells and remove the big outputs (ie for importing libraries, etc) from the data publishing notebook?

Also, where there anything you changed in the data publishing notebook that is supposed to be there? I didn't see anything but it's hard to tell without a proper diff. 

The last thing will be to update the Foundry documentation and example notebooks with the syntax change before we can cut this release.

Issue #187, opened 2022-06-08
Update `publish()` to take links


kjschmidt913:
Need to take "links" to pass onto the MDF connect_client. To do this one, need to add helper function to MDF connect client first (something like add_links()

- [x] add helper functions to Connect Client
- [ ] add logic in Foundry publish() to pass on links metadata

before this story also included passing in all datacite data. preserving that text here in case it's needed in the future:

need to take all necessary keys to add to the datacite block (called dc) (e.g. description, resource_type, etc)
For "description", in data cite it's just Description, and then you can give it a DescriptionType:“Abstract”. Link to dc metadata schema: https://github.com/materials-data-facility/data-schemas/blob/12bb3c7da2a5d8667e5cdb8fe90f18b77094af4d/schemas/dc.json#L411

So it'd be something like:
```
{
    "dc": {
        "descriptions": [{
            "description": "Blah blah blah",
            "descriptionType": "Abstract"
        }]
    }
}
```
Things I need in the dc block:

```
def create_dc_block(self, title, authors,
                        affiliations=None, publisher=None, publication_year=None,
                        resource_type=None, description=None, dataset_doi=None,
                        related_dois=None, subjects=None,
                        **kwargs):
```

kjschmidt913:
dependent on issue #23 being completed first

blaiszik:
Check Foundry Miro board for metadata format.

Some ideas

```"links": [{
 "type": "dataset_paper",
   "doi": "10.1072/xyz",
 "bibtex": "<bibtex text>",
   "url": "https://doi.org/10.1072/xyz"
  },
"type": "other_paper",
 "doi": "10.1072/xyz",
 "bibtex": "<bibtex text>",
   "url": "https://doi.org/10.1072/xyz"
  },
  {
   "type": "model",
   "doi": "10.1072/thy",
   "bibtex": "<bibtex text>"
   "url": "https://doi.org/10.1072/thy"
  },
  {
   "type": "code",
   "doi": "10.1072/thy",
   "<bibtex text>"
   "url": "https://doi.org/10.1072/thy"
  },
 {
 "type": "dataset",
   "doi": "10.1072/thy",
   "<bibtex text>"
   "url": "https://doi.org/10.1072/thy"
  }
 ]
```

```
(this)dataset_paper: The paper where the dataset was released
other_paper: Other papers associated with or using the dataset
model: Links to models
code: Links to code
```

ascourtas:
[...]

PR #390, opened 2023-08-01
Publish links

blue442:
Addressing #187 by adding the functionality where `foundry.publish_dataset()` can take links in the form of a keyword argument. The arguments associated with the `links` keyword are of the following form:
`{"type":str, "doi":str, "url":str, "description":str, "bibtex":str}`
following the convention laid out in the [MDF data schemas](https://github.com/materials-data-facility/data-schemas/blob/master/schemas/links.json), and can be included as a single argument or a list of arguments.

Some light validation is performed to ensure that the keys in each of the link conform to the options described in the MDF schema (but are all optional); violation results in a `ValueError` being thrown. 

After validation the links are passed to the `add_links()` function from the MDF connect client.

Tests were included in `test_foundry.py` to test the successful addition of links as well as ensuring the validation process works successfully.


ascourtas (changes requested):
Looks great! All that's left to finalize this is updates in the GitBook documentation and in the example notebooks (under `/examples`). Let me know if you're having trouble with the GitBook documentation -- you can open an Edit Request which is like a PR there

PR #406, opened 2023-10-30
remove breaking changes that load on init

ascourtas:
Fix up `main` to revert load on init() changes previously merged, but keep flake8 fixes, comments, docs, metadata validation, and minor tweaks. This means we preserve previous functionality for users when loading datasets, but keep all other changes.

This should keep `main` "clean" such that we can cut a minor/patch release before the new code-breaking changes of the class refactor are merged.

### DIFF
--- a/foundry/foundry.py
+++ b/foundry/foundry.py
@@ -50,27 +50,17 @@ class Foundry(FoundryBase):
     auths: Any
 
     def __init__(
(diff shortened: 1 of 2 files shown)


blue442 (approved):
Checked diffs for a couple different commits, and it looks like all of the target functionality was removed while leaving the rest - LGTM!

